PR #57, opened 2012-07-19
Add support for MapBox hosting compositing

ansis:
Now, by default, consecutive MapBox tile layers composited together. Disabled layers do not interrupt compositing groupings.

Compositing can be disabled on a layer by calling: `layer.composite(false)`


ansis:
Todo: Does draw need to be called after layers are added or removed? 
EDIT: taken care of in e647540

Issue #244, opened 2013-02-28
mapbox.map proxy?


tmcw:
Should we alias L.Map to mapbox.map?

jfirebaugh:
I think we should rename layerGroup to map. PR in a sec.

PR #245, opened 2013-02-28
layerGroup -> map, dataLayer -> markerLayer

jfirebaugh:
The layerGroup rename + removing auto seems like a good solution to the callback confusion. In situations where a callback is needed, the map 'load' event can be used instead.

I'm convinced about the dataLayer -> markerLayer rename. The key factor was realizing that the target of mapbox.js is the MapBox API, not a generic TileJSON implementation.

The examples cleaned up very nicely.

Ref #244.


tmcw:
I like the direction of this, only a few issues:
- Need to figure out how multiple layers should be represented - should there be a shortcut?
- `mapbox.map` isn't documented yet

jfirebaugh:
Docs added.

There's a couple ways to do multiple layers:
- Compositing -- pass comma-separated ids.
- Use `mapbox.map` as the "base layer" and then add additional `mapbox.tileLayer`s (example: opacity.html).

Did you have a third way in mind?

Issue #317, opened 2013-04-11
map is limited to zoom 18


ansis:
Maps shouldn't be limited to 18 by default.

ansis:
The same cause as #308 

Leaflet assumes that most layer options won't change after being loaded. Is it worth trying to get this changed in Leaflet, or should mapbox.map only add the layers after they are ready, as Leaflet expects? Is there any point to adding a layer before its loaded?

jfirebaugh:
The reason we add layers before they are loaded is that leaflet has limited ability to reorder layers, and added layers should have ordering based on when they were added, not when their loads happened to complete. Though maybe we can do something with `TileLayer#setZIndex()`.

ansis:
Fixed by #331

PR #331, opened 2013-04-12
update calculated map properties on layer change

### LINKED ISSUE #308: setView changes attribution status
Diagnosis is: only when the map is 'ready' does it check to see if there are any layers that get `onAdd` called on them and trigger `layeradd`, and `layeradd` triggers an Attribution change.

ansis:
This recalculates zoom bounds and attribution when layers have changed (map is finished initializing, or 'ready' is called on a layer). Does this seem reasonably nice?

This would fix #308 and #317

### DIFF
--- a/src/grid_layer.js
+++ b/src/grid_layer.js
@@ -10,8 +10,6 @@ var GridLayer = L.Class.extend({
     includes: L.Mixin.Events,
 
     options: {
-        minZoom: 0,
-        maxZoom: 18,
         template: function() { return ''; }
     },
 
--- a/src/map.js
+++ b/src/map.js
@@ -59,6 +59,12 @@ var Map = L.Map.extend({
         }
     },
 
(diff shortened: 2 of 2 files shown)


tmcw:
Hmm, if you add a layer and call `loadID` on it, will it fire `ready` twice and add a new attribution in addition to its old one?

ansis:
Yeah, it doesn't remove old attribution. We'd need to have a separate 'change' event that gets fired with the old options or change the way L.AttributionControl works.

If a layer is created and only loaded once it works fine. Its only a problem if you reuse a Layer for multiple IDs. This sort of goes against the idea of how Leaflet layers work.

Do we want to support this kind of in-place layer mutation?

tmcw:
> Do we want to support this kind of in-place layer mutation?

We could remove the loadID and loadURL apis if we don't support this. Maybe it leads to a simpler api by being more limiting?

PR #401, opened 2013-05-25
Update JS/CSS URLs in documentation

stuartpb:
Version bump to 1.0.2
Use protocol-independent URLs

### DIFF
--- a/API.md
+++ b/API.md
@@ -28,11 +28,11 @@ Here's a simple page that you can set up with MapBox.js:
 
     <html>
     <head>
-      <link href='http://api.tiles.mapbox.com/mapbox.js/v1.0.0/mapbox.css' rel='stylesheet' />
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Hey, thanks for the pull request.

Our SSL support is still in [private beta](http://www.mapbox.com/developers/ssl-api/), and we ask customers to contact us before using it, so we don't want to reference the cloudfront.net URLs by default in our documentation.

I can take the version change part if you separate that out -- and the most recent version is 1.0.3.

tmcw:
Closing, will manually make a version upgrade.

stuartpb:
SSL support isn't actually activated unless you set `L.mapbox.config.HTTPS_URLS`, though. Using the cloudfront.net URL only changes the browser's security response to the exact same resources: with it, HTTPS pages don't fail to load the script (the script still requests all the tiles over non-SSL HTTP, so no beta functionality is touched).

tmcw:
Access to the javascript API over SSL is another part of the beta. This is covered in our documentation for the private beta, not publicly and not in these docs.

PR #409, opened 2013-06-04
Documented an undocumented option.

iammerrick:
Added documentation for a useful option.


tmcw:
Thanks! Though, in the prose above those docs is

> The third argument is optional. If provided, it is the same options as provided to L.Map with the following additions:

`zoomControl` is [one of the options in L.Map](http://leafletjs.com/reference.html#map-options) - we aren't re-documenting it here.

iammerrick:
Got it! Didn't see that before! Thank you!

PR #419, opened 2013-06-15
Fix link to API documentation in README.md.

lchski:
Was missing the http protocol stub at the beginning to make it a non-relative URL.


tmcw:
Thanks!

PR #431, opened 2013-06-26
Simple SSL sketch. Needs testing on production.

tmcw:
This needs testing, but I think it'll work. The SSL detection line is taken from how Google Analytics does it, so it's pretty well-tested.

### DIFF
--- a/src/url.js
+++ b/src/url.js
@@ -8,42 +8,28 @@ var config = require('./config');
 // against multiple `CNAME`s in order to avoid connection limits in browsers
 module.exports = {
     base: function(hash) {
+        var isSSL = 'https:' === document.location.protocol;
         // By default, use public HTTP urls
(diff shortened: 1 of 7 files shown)


src/url.js
```diff
@@ -8,42 +8,28 @@ var config = require('./config');
...
+        var urls = (isSSL || config.FORCE_HTTPS) ?
+            config.HTTPS_URLS : config.HTTP_URLS;
 
         if (hash === undefined || typeof hash !== 'number') {
             return urls[0];
         } else {
             return urls[hash % urls.length];
         }
-    }, 
-    httpsify: function(tilejson) {
```
jfirebaugh, line 27: This is used in a couple places.

tmcw, line 27: yep, have another commit that fixes and adds tests.

PR #449, opened 2013-07-10
Redesign geocoder to echo v1 style

tmcw:
Needs
- [ ] Icon
- [ ] Results list if multiple are found?
- [ ] Style simplification and tweaks


tmcw:
Made more progress here, but bus wifi blocks git.

tmcw:
Fixed in 6d5726c5cbe1c6912288185ff895aa56b1f124bf

PR #498, opened 2013-10-03
add browserify instructions

zeke:
Auto-attaching to `window.L` behavior seems unconventional to me from an npm philosophy perspective: I'd rather assign it myself:

``` js
window.L = require('mapbox.js');
```

But rather than digging into really solving that, I'm offering a bit of documentation for others who may have hit the same wall I did.


max-mapper:
+1 keep your grubby hands off my globals

tmcw:
Yes, not sure whether there was some historical reason for the way we're doing things (the browserify `-s` flag had some issues) and whether they're undoable now...

tmcw:
Thanks! Merged.

PR #502, opened 2013-10-09
Marker -> polyline example

bsudekum:
Just for fun.

cc @tmcw

### DIFF
--- a/_posts/examples/v1.0.0/0100-01-01-line-marker.html
+++ b/_posts/examples/v1.0.0/0100-01-01-line-marker.html
@@ -13,20 +13,18 @@ tags:
 var map = L.mapbox.map('map', 'examples.map-zr0njcqy');
 
 map.markerLayer.on('ready', function(e) {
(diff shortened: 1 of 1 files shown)


_posts/examples/v1.0.0/0100-01-01-line-marker.html
```diff
@@ -0,0 +1,32 @@
...
+<script>
+var map = L.mapbox.map('map', 'examples.map-zr0njcqy');
+
+map.markerLayer.on('ready', function(e) {
+    var markers = [];
+    var line = [];
+
+    this.eachLayer(function(marker) {markers.push(marker); });
+
+    for(i=0;i<markers.length;i++){
```
tmcw: Please use [airbnb style](https://github.com/airbnb/javascript), so spaces between these things, and also `for(i=0` creates a global `i` var - it should be `for (var i = 0; i < markers.length; i++) {`

PR #504, opened 2013-10-10
fix for retina markers

samanpwbb:
This fix forces the background image size of markers to exactly match  the size of the container. This will work as long as our marker containers match the size of the intended background image size, which they all do now.

### DIFF
--- a/theme/style.css
+++ b/theme/style.css
@@ -178,6 +178,11 @@ a.mapbox-icon-twitter {
     padding: 5px 8px;
 }
 
+/* Fix for retina markers */
+.leaflet-marker-icon {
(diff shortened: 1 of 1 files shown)


tmcw:
Hmm this seems wrong - we try to very very minimally change stock Leaflet behavior here, and I suspect this could regress user styles.

samanpwbb:
@tmcw there is no way to specifically select MapBox markers. How did we do retina markers in the past? Have these ever worked?

samanpwbb:
Closing, this dealt with a weird edge case where marker images were being made with background-image rather than img tags.

PR #530, opened 2013-11-05
table out the api docs

tristen:
This commit removes the overview section and creates tables for each argument to functions.

### DIFF
--- a/API.md
+++ b/API.md
@@ -180,7 +180,7 @@ Load GeoJSON data for this layer from the URL given by `url`.
 
 _Example_:
 
-    var markerLayer = L.mapbox.markerLayer(geojson)
+    var markerLayer = L.mapbox.markerLayer()
         .addTo(map);
 
     markerLayer.loadURL('my_local_markers.geojson');
@@ -189,7 +189,7 @@ _Returns_: the layer object
 
 ### markerLayer.loadID(id)
 
-Load tiles from a map with the given `id` on MapBox.
+Load marker GeoJSON data from a map with the given `id` on MapBox.
 
 | Options | Value | Description |
 | ---- | ---- | ---- |
@@ -197,7 +197,7 @@ Load tiles from a map with the given `id` on MapBox.
 
 _Example_:
 
-    var markerLayer = L.mapbox.markerLayer(geojson)
+    var markerLayer = L.mapbox.markerLayer()
         .addTo(map);
 
     // loads markers from the map `examples.map-0l53fhk2` on MapBox,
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -386,42 +141,27 @@ _Example_:
...
-
-* `filter`: A function that accepts a feature object and returns `true` or `false`
-  to indicate whether it should be displayed on the map. This can be changed
-  later using `setFilter`.
-* `sanitizer`: A function that accepts a string containing tooltip data, and returns a
-  sanitized result for HTML display. The default will remove dangerous script content,
-  and is recommended.
+| Options | Value | Description |
+| ---- | ---- | ---- |
+| id _or_ url _or_ tilejson | string | Must be either <ul><li>An id string examples.map-foo</li><li>A URL to TileJSON, like `http://a.tiles.mapbox.com/v3/examples.map-0l53fhk2.json`</li><li>A GeoJSON object, from your own Javascript code</li><li>`null`, if you wish to only provide `options` and not initial data.</li></ul> |
```
tmcw: In a case like this, we need to preserve the detail that was before: if you give a tilejson argument, it is not a string.

API.md
```diff
@@ -430,34 +170,34 @@ _Example_:
...
-
-1. `string` a URL referencing a GeoJSON resource
+| Options | Value | Description |
+| ---- | ---- | ---- |
+| url | string | A map id |
 
 _Example_:
 
-    var markerLayer = L.mapbox.markerLayer()
+    var markerLayer = L.mapbox.markerLayer(geojson)
```
tmcw: This is a regression - this example should not both specify a geojson object and load a url of geojson.

API.md
```diff
@@ -430,34 +170,34 @@ _Example_:
...
 
     markerLayer.loadURL('my_local_markers.geojson');
 
 _Returns_: the layer object
 
-### markerLayer.loadID(id: string)
+### markerLayer.loadID(id)
 
-Load marker GeoJSON data from a map with the given `id` on MapBox.
+Load tiles from a map with the given `id` on MapBox.
```
tmcw: Also a regression: this function does not load tiles.

API.md
```diff
@@ -430,34 +170,34 @@ _Example_:
...
-
-1. `string` a map id
+| Options | Value | Description |
+| ---- | ---- | ---- |
+| url (_required_) | string | A map id |
 
 _Example_:
 
-    var markerLayer = L.mapbox.markerLayer()
+    var markerLayer = L.mapbox.markerLayer(geojson)
```
tmcw: Same as before.

API.md
```diff
@@ -817,16 +491,18 @@ A `L.Marker` object with the latitude, longitude position and a styled marker
 
 # Utility
 
-## L.mapbox.sanitize(text: string)
```
tmcw, line 820: Why doesn't this have an Arguments section?

mourner:
Amazing work on the docs. Looks much better now.

PR #535, opened 2013-11-07
adds ridelapse example

geografa:
@tmcw This seems a bit much for an example. If you have ideas for slimming it down...


tmcw:
Added code style docs: https://github.com/mapbox/mapbox.js/blob/master/EXAMPLES.md#code-style - please reformat the code to taste
- `var coor = ride.features[0].geometry.coordinates;` - why?
- Needs plenty of explanatory code comments
- This code looks to me like it will run off the end of the coordinates array when `i` gets higher than the number of bits in the line segment. @rsudekum is that the case?

tmcw:
Closing; https://github.com/mapbox/mapbox.js/blob/gh-pages/_posts/examples/v1.0.0/0100-01-01-marker-movement.html is hitting this

PR #539, opened 2013-11-11
Title for each example, viewport scale.

bsudekum:
Viewport scale got lost in recent updates. This PR also adds the title of example to each example.

cc @tmcw


tmcw:
Cool, green button it.

Issue #485, opened 2013-08-29
Redesign attribution control


heyitsgarrett:
_[Internal MapBox context here](https://github.com/mapbox/frontend/issues/172)_

As part of our work with promoting the relationship between OSM and MapBox maps, I built a lightweight Leaflet plugin ([demo](http://www.mapbox.com/labs/attribution)) that combined attribution and an 'Improve this map' link.

![cap](https://f.cloud.github.com/assets/200357/1050950/70c0fe40-10c2-11e3-8725-cf70cda6d916.gif)

This style could be refactored nicely as the default attribution for mapbox.js, with an optional 'improveMapLink' boolean attribute.

incanus:
This button borrows heavily from the iOS look, which doesn't look like that anymore on iOS 7. 

![6-7](https://f.cloud.github.com/assets/17722/1077335/546c80e8-14fa-11e3-8eb1-fb5fa8aafc56.png)

heyitsgarrett:
Would like to consider this for mapbox.js 1.5, including thinking about how the above 'Improve this map' interaction ideas could play a part in connecting dots between consumers of the map and OSM contributors.

tmcw:
@heyitsgarrett the time is now: Leaflet v0.7.0 is launched and work is happening in the v1.5.0 branch. Would love to see if you can knock this out this week?

tristen:
Taking this one on in https://github.com/mapbox/mapbox.js/tree/mapbox-attribution-control

heyitsgarrett:
@tristen is a super:star:

tristen:
Merged!

PR #563, opened 2013-11-20
Mapbox info control

tristen:
This commit adds a new info control in reference to #485 
- [x] Add and remove attribution on layerAdd and layerRemove like [in L.Control.Attribution](https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Attribution.js#L27-L29)
- [x] IE test

I stuck with an _info_ icon as this felt visually stronger as an icon as is more diverse than a &copy; I've also placed this control in the bottom left corner (as opposed to the bottom right) as it created some awkwardness when a legend was just above it.

Here's a little animation of it. The control appends any attribution content passed in to `L.mapbox.infoControl.addInfo()` and adds an "Improve this Map" link by default. There's also a supported removeInfo method. 

attribution


incanus:
Was gonna say: in addition to any other reasons, the edit in OSM link could be really tricky on mobile web devices, where I don't think iD is meant to tread just yet...

tmcw:
From meatspace: we are re-adding edit in osm, making it off by default, and showing it when (requested && !mobile)

tristen:
@incanus right - I'm conditioning the display of this on mobile with `!L.Browser.mobile`

src/info_control.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+
+var InfoControl = L.Control.extend({
+    options: {
+        position: 'bottomleft',
+        sanitizer: require('sanitize-caja'),
+        editinosm: false
```
jfirebaugh: Options names are camel case, suggest `editLink`.

tristen: Sounds good

src/info_control.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return this;
+    },
+
+    _osmlink: function() {
+        var center = this._map.getCenter();
+        var z = this._map.getZoom();
+        window.open('http://www.openstreetmap.org/edit?' + 'zoom=' + z +
+        '&editor=id' + '&lat=' + center.lat + '&lon=' + center.lng);
```
jfirebaugh: Don't include `editor=id` here, the default is to use an appropriate editor automatically (iD for most users, preferred editor for users who have set one, P2 for IE users).

src/map.js
```diff
@@ -15,7 +16,9 @@ var Map = L.Map.extend({
         markerLayer: {},
         gridLayer: {},
         legendControl: {},
-        gridControl: {}
+        gridControl: {},
+        infoControl: {},
+        attributionControl: false
```
jfirebaugh: Should L.mapbox.map _not_ create an infoControl by default if someone explicitly passes `{attributionControl: true}`?

PR #578, opened 2013-11-26
Add Mobile subsections by fixing an omission in the generator script.

tmcw:
Is there a way to specify a different thing for the title versus the
link? Probably, but I am running away from YAML as quickly as possible.

/cc @tristen


tristen:
Holy radar dropping on my part. Thankfully this is all linked up now thanks to @tmcw

PR #600, opened 2013-12-12
better info icon on gray maps

samanpwbb:
Solves problem where info icon is illegible on gray maps.

Old:
screen shot 2013-12-12 at 5 31 01 pm
screen shot 2013-12-12 at 5 36 07 pm

New:
screen shot 2013-12-12 at 5 34 35 pm
screen shot 2013-12-12 at 5 36 31 pm


tmcw:
What's up with removing the 404040 color build? Was it unused?

samanpwbb:
@tmcw i replaced it with 000000

PR #607, opened 2013-12-18
Include simplestyle.style in API

tmcw:
@jfirebaugh could you review this?

Behavior of `map` constructor is:
- If `markerLayer` is provided, it pre-empts `featureLayer`, otherwise `featureLayer` is the default.
- `map.markerLayer` and `map.featureLayer` point to the same thing
- [ ] Add a deprecation notice to the API


test/spec/feature_layer.js
```diff
@@ -0,0 +1,159 @@
+describe('L.mapbox.markerLayer', function() {
```
jfirebaugh: Looks like the content of this file and `test/spec/marker_layer.js` are swapped.

jfirebaugh:
Looks good.

> Behavior of map constructor is...

Let's add a test for this.

Anything we need to consider from `marker.js`, like making `createPopup` feature-generic?

tmcw:
> Anything we need to consider from marker.js, like making createPopup feature-generic?

I can't think of anything, but will test polygon-popups before shipping.

Fixed up the test comment, renamed, and tested the options aliasing.

jfirebaugh:
:shipit:

PR #610, opened 2013-12-24
Remove trailing comma from bower.json

benbalter:
Otherwise, bower can't parse the JSON.

//cc @skalnik


skalnik:
Woops, my bad. Nice catch @benbalter!

PR #615, opened 2014-01-06
Integrate Leaflet docs on page, add 1.6.0 docs

tmcw:
This integrates Leaflet docs inline with mapbox.js docs.
- How do you pull a specific version of Leaflet docs? doesn't seem like we can use the npm package or git tags. /cc @mourner 
- This makes the page extremely long, too long in my opinion for actual use. I think this needs more structure like... collapsible sections, or mapbox.js & leaflet on different pages, or something else? Thoughts @tristen ?
- Leaflet's `reference.html` is not one thing - it's a TOC, an intro, and Javascript code. I've removed this manually for the time being, and could write some javascript `.replace()` to do the same thing, but this seems icky and hacky and I would like not to. /cc @mourner


mourner:
Yeah, I wasn't tagging the website, just keeping `gh-pages` and `gh-pages-0.x` (deleting older branches).

tmcw:
Okay, pushed one possibility, would love thoughts on it. There are still gotchas like:
- We style all things with `#map` as big maps anywhere they are regardless of anything. LOL?
- If Jekyll was ever good, it ain't now. Using Jekyll for docs sucks. I want to not.
- Leaflet docs don't work with links yet.

Right now IA is

Landing page (with intro docs) -> (all | individual page)

tmcw:
Bonus: our approach to links doesn't scale at all. Links should not be the way they are, they should be autodetected or generated post-facto.

Leaflet being in html and mapbox being in md sucks.

PR #629, opened 2014-01-14
Fixed layout

tmcw:
- [x] We need a page to link to to show examples
- [x] We need a page to link to to show plugins
- [x] We need to add an obvious link back to mapbox. This means a logo like the rest of the site, not a hexagon or a text link.
- [x] Framing for plugins landing page
- [x] Framing for examples landing page


tmcw:
- This should have a MapBox logo in the top left. It's cool that it's different from the rest of the site, but this is not part of the 'app' and for a vast majority of users it will be a blank user image, which does not say "go to the home page"
- The plugins page cannot be a sidebar: it needs to be a webpage, so it needs a URL that people can link to. Let's move plugins to the right hand side so that they can have content and not all URLs are clipped.
- Plugins need to have license text
- TOC sections do not highlight active links, and need to.

aaronlidman:
@tristen 
- here are the changes I brought up https://github.com/mapbox/mapbox.js/tree/tighterTOC
- take it or leave it but I think it makes things clearer
- not really feeling the examples thumbnails, many of the concepts there are not simple enough to make into an image
- I'm working on plugins now

screen shot 2014-01-16 at 12 53 59 pm

tmcw:
I agree with Aaron here, thumbnails just aren't effective for examples right now. Can we get a `/plugins` homepage back? Would love to have something simple and scrollable.

tristen:
> Thumbnails just aren't effective for examples right now

I think its worth investing time in the examples: Adding great thumbnails, Having a collection of hot base maps to use for them, improving any example UI, and writing better descriptions. In the meantime:

screenshot_2014-01-16screen shot 2014-01-16 at 6 24 04 pm

I still see this flat list in descending order being a valuable layout listing right now.

tmcw:
When we have great thumbnails, we should add design to accentuate them, if possible. Until then, it does not make sense to have parts of the website that will make sense in the future: people do not know they will be great, only that they aren't great right now.

I would love to refocus on getting this straightforward and simple rather than rolling in new design experiments.

tristen:
Yep I hear you. Will break the thumbnail work into a new branch.

tmcw:
We really need more than a hexagon here. Sorry to be so naggy about this, but we need to stop trading clarity for brevity.

lxbarth:
Echoing @tmcw, the /mapbox.js proposed here doesn't tie back to Mapbox.com.

What are our goals for redesigning mapbox.js?

Without context, here are additional problems I find with navigation and information hierarchy:
- Can't link to Examples or Plugins page (user story: link drop mapbox.js examples to novice to get started)
- Clicking on Examples/API/Plugins.js doesn't take me to a landing page
- I need to click on Mapbox.js v1.6.0 to get to the landing page of the API
- There should be a simple beautiful landing page for each section, framing expectations. There is only one for API

lxbarth:
To illustrate this more. This is the current [mapbox.js examples landing page](https://www.mapbox.com/mapbox.js/example/v1.0.0/). It is simple and effective:
- Start concrete with the most simple example
- Clear actions: use the simple example, or look for others
- Has a URL. This is one of the top three pages I link drop.

screen shot 2014-01-16 at 5 44 53 pm

aaronlidman:
For me:
- [x] Better plugin layout, it's haphazard and awkward right now
- [x] Link to origin of each plugin (currently coupled with license)
- [x] Link to full documentation

tmcw:
@aaronlidman in the interest of getting this to mergetown faster, feel free to punt on adding any more plugins and any screenshots of them.

@tristen 



Can we please stop circling around the simple solution here and getting caught up in this distraction?

Adding a hit-list of launch tasks to the PR description.

tristen:
> Can we please stop circling around the simple solution here and getting caught up in this distraction?

Please outline the problem here.

tmcw:
- If users can't find the homepage link of the website, they will type in `mapbox.com` again in the URL bar. This is not a theoretical problem, this is exactly what happened with the spaceship redesign, and it hit even internet-savvy users like former Mapbox employees. We should not make this mistake again.
- Mapbox.js is a common landing page, even from outside of the site, via links. Why would we give up our ability to brand here, and to offer a clear way to say "oh, this is a cool api, where is it from? let me buy this thing"

On the other hand:
- Why would we spend this much time trying to reinvent the top left corner of the page when we (1) already have an easy, simple solution and (2) have actual bugs to fix.

tmcw:
Okay, let's land this. Let me know if I can pick up tasks, I'm ready to tag & deploy mapbox.js v1.6.1 as soon as @tristen is happy with the site. @aaronlidman is writing a blog post about the plugins page and I'll write a post about the API docs & mapbox.js update

Issue #640, opened 2014-01-21
Passing TileLayer options to L.mapbox.Map


rclark:
The docs are a little ambiguous / don't actually tell you that you can pass TileLayer options when you're creating a map.

![screen shot 2014-01-20 at 5 55 39 pm](https://f.cloud.github.com/assets/515424/1959784/acb23fcc-8237-11e3-818a-6b71a11785bf.png)

Unclear as evidenced by my attempts to support retina displays...

``` javascript
var map = L.mapbox.map('map', 'rclark.map-hibbity', { detectRetina: true });
```

which didn't work so...

``` javascript
var map = L.mapbox.map('map', 'rclark.map-hibbity', { tileLayer: false });
L.mapbox.tileLayer('rclark.map-hibbity', { detectRetina: true });
```

which seemed redundant.

Then I [read the code](https://github.com/mapbox/mapbox.js/blob/master/src/map.js#L34-L37) and found out that you can just

``` javascript
var map = L.mapbox.map('map', 'rclark.map-hibbity', { tileLayer: { detectRetina: true } });
```

I'm happy to take a stab at adjusting the docs if it seems useful to others.

tmcw:
Definitely could be clearer here: take a shot at improving this & PR it.

rclark:
Related things that came up = 
- there are other options for any of the different controls / layers that could be defined by TileJSON. Worth documenting?
- I didn't find any developer docs about how to build docs locally to see your changes. This is what I did, but maybe there's a better way or I missed something? I'd like to write some instructions into a readme if that's okay.

``` sh
git clone git@github.com:mapbox/mapbox.js && cd mapbox.js
# ... make changes ...
git checkout -b update-docs
git commit -am "changed API.MD file"
npm link
cd .. && git clone git@github.com:mapbox.js mapbox.js.docs -b gh-pages
cd mapbox.js.docs/_docs
npm install
npm link mapbox.js
./build.sh
cd .. && jekyll serve
```

tmcw:
Can you add more detail to documentation at https://github.com/mapbox/mapbox.js/blob/master/DOCUMENTING.md ?

Also, if you're going to advise an `npm link` approach please be super explicit that the result of that should never be deployed: docs are deployed off of tagged versions, and never updated in place unless absolutely necessary.

rclark:
> Also, if you're going to advise an npm link approach

I'll write what I did up in that file, but if you have a smoother approach I'm all ears.

[...]

PR #641, opened 2014-01-21
Clarify map options

rclark:
Parameters passed to `L.mapbox.map` options are passed on to the child layers defined in TileJSON. Just trying to elaborate that in the docs. See #640


tmcw:
Great!

PR #654, opened 2014-01-23
Fixed broken link from L.mapbox.map page to L.map

lyzidiamond:
Was linking to .../l-map, but actual page is .../l-map-class!


mourner:
@jfirebaugh forgot to comment?

jfirebaugh:
Huh, no: apparently, when you delete a branch that a pull-request is targeting, it gets closed automatically without comment. Sorry about that @lyzidiamond -- we're doing some reshuffling behind the scenes and the docs branch is now `mb-pages`.

That said, I was meaning to comment on this pull request anyway. The file you updated is auto-generated via [a script](https://github.com/mapbox/mapbox.js/blob/mb-pages/_docs/generate.js), so ideally this should be addressed there so that future versions don't have the same broken link. I'll open an issue on this.

lyzidiamond:
Noted!

PR #659, opened 2014-01-28
make marker popups work better

samanpwbb:
- Added overflow-auto and max height to popups
- Fixed overly broad and aggressive rule intended for tiles that was interfering with image styles in popups
- Made base type styles smaller

From:
screen shot 2014-01-28 at 3 26 18 pm

To:
screen shot 2014-01-28 at 3 26 02 pm

### DIFF
--- a/theme/style.css
+++ b/theme/style.css
@@ -757,9 +757,9 @@
 /* Browser Fixes
 ------------------------------------------------------- */
 /* Map is broken in FF if you have max-width: 100% on tiles */
-.leaflet-container img.leaflet-tile { max-width:none; }
+.leaflet-container img.leaflet-tile { max-width:none !important; }
 /* Stupid Android 2 doesn't understand "max-width: none" properly */
-.leaflet-container img.leaflet-image-layer { max-width:15000px; }
+.leaflet-container img.leaflet-image-layer { max-width:15000px !important; }
 /* workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=888319 */
 .leaflet-overlay-pane svg { -moz-user-select:none; }
 /* Older IEs don't support the translateY property for display animation */
--- a/theme/style.css
+++ b/theme/style.css
@@ -1,13 +1,14 @@
 /* general typography */
 .leaflet-container {
   background:#fff;
-  font:15px/25px 'Helvetica Neue', Arial, Helvetica, sans-serif;
+  font:12px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
   color:#404040;
   color:rgba(0,0,0,0.75);
   outline:0;
   overflow:hidden;
   -ms-touch-action:none;
   }
+
 .leaflet-container *,
 .leaflet-container *:after,
 .leaflet-container *:before {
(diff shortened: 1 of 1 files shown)


tmcw:
Hm, looks good but enforcing a max-height in CSS may be problematic: [Leaflet writes its styles from JS-land](http://leafletjs.com/reference.html#popup) and users with custom maxHeight values will see changed behavior. /cc @mourner

samanpwbb:
I fixed this on mapbox.com side as a short term solution: https://github.com/mapbox/mapbox.github.com/commit/0444bf3076031a3d7633c628d61094719c0169ca

mourner:
@samanpwbb why not just use `maxHeight` option of Leaflet?

samanpwbb:
I'd like this sort of thing to be handled by CSS. 

I'm guessing Leaflet's `maxHeight` sets an inline attribute on the element? In that case, `maxHeight` will override the CSS rule, so I don't think having this in the distributed CSS file is a problem.

theme/style.css
```diff
@@ -746,9 +757,9 @@
 /* Browser Fixes
 ------------------------------------------------------- */
 /* Map is broken in FF if you have max-width: 100% on tiles */
-.leaflet-container img { max-width:none!important; }
+.leaflet-container img.leaflet-tile { max-width:none; }
```
mourner: The reason why I had to put important here is that some stupid CSS frameworks like Bootstrap use `!important` too, which will break Leaflet maps unless overriden with `!important` from Leaflet side. E.g. https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.css#L222
So I'd suggest not touching these.

mourner:
One reason it was not entirely handled in CSS is that I wanted to add a special class when popup reaches max height so that I could style the scrolled container, e.g. add borders like Leaflet does now.

samanpwbb:
Okay, reverted to importants. Now that the rule applies to `img.leaflet-tile` rather than `img` this shouldn't interfere with too much.

theme/style.css
samanpwbb: That is the silliest rule I've ever seen. As much as I'd like to step out of this CSS specificity arms race, I don't see how your override could hurt as long as it applies only to tiles.

samanpwbb:
Removed the max height rule with bc3b5fe49a94621cc5a32af467def95065c759b5 so this should now be good to go. Sounds like it's best to leave max height management to the front end.

theme/style.css
mourner: Right. I should be more specific in Leaflet as well.

samanpwbb:
@tmcw / @mourner I think this should be fine now but I await approval.

theme/style.css
```diff
@@ -24,9 +25,14 @@
...
 .leaflet-container p {
   font-size:15px;
-  line-height:25px;
+  line-height:20px;
   margin:0 0 10px;
   }
+
+.leaflet-container img {
+  margin-bottom: 10px;
+  }
```
mourner, line 34: shouldn't this be more specific?

mourner:
Otherwise looks good.

tmcw:
Y, we should not style all `.leaflet-container img` with a margin-bottom.

samanpwbb:
Cool, I made the style more specific. I am ok totally cutting it as well if you guys still feel weird about it. I think it's ok to be a little opinionated with styles within the `marker-description` though - this is in 95% of the cases just going to make markers look better for people. The other 5% are people who are already messing with popup appearance and then it's a one css line override.

mourner:
Yeah, looks good now.

PR #660, opened 2014-01-29
Compress CSS

tmcw:
This reduces CSS weight from 20KB to 16KB by running clean-css over the
file. I think this is worth it, since it doesn't introduce much of a
development overhead.

As a side-effect, this would introduce mapbox.uncompressed.css, in line
with mapbox.uncompressed.js.

@tristen what do you think? CSS compression, maybe the future?


mourner:
@tmcw just wondering, what's the difference when gzipped?

tmcw:
3131 bytes compressed & `gzip -9f`, 3875 bytes just gzipped.

PR #662, opened 2014-01-30
Improve this map

tristen:
This commit:
- Surfaces the `editlink` option in infoControl to be on by default
- Adds an `open` option to have attribution turned on by default
- Style, "Improve this map" link. Bump up font size and weight and drop to a new line.

References https://github.com/mapbox/www.mapbox.com/issues/3239
#### TODO
- [x] Finalize style
- [x] Fix tests

### DIFF
--- a/src/info_control.js
+++ b/src/info_control.js
@@ -98,12 +98,12 @@ var InfoControl = L.Control.extend({
         this._content.innerHTML += info.join(' | ');
 
         if (this.options.editLink && !L.Browser.mobile) {
-            this._content.innerHTML += (info.length) ? ' | ' : '';
             var edit = L.DomUtil.create('a', '', this._content);
             edit.href = '#';
+            edit.className = 'mapbox-improve-link';
             edit.innerHTML = 'Improve this map';
-            edit.title = 'Edit in OpenStreetMap';
-            L.DomEvent.on(edit, 'click', L.bind(this._osmlink, this), this);
+            edit.title = 'Improve this map';
+            L.DomEvent.on(edit, 'click', L.bind(this._editlink, this), this);
         }
 
         // If there are no results in _info then hide this.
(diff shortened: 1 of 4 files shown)


lxbarth:
This is great. This is done from my perspective. @tmcw - want to give this a review and merge?

src/info_control.js
```diff
@@ -106,11 +111,13 @@ var InfoControl = L.Control.extend({
...
-    _osmlink: function() {
+    _editlink: function() {
+        var tilejson = this._tilejson || this._map._tilejson || {};
+        var id = tilejson.id || '';
         var center = this._map.getCenter();
         var z = this._map.getZoom();
-        window.open('http://www.openstreetmap.org/edit?' + 'zoom=' + z +
-        '&lat=' + center.lat + '&lon=' + center.lng);
+        var url = 'https://www.mapbox.com/map-feedback/#' + id + '/' + center.lng + '/' + center.lat + '/' + z;
+        window.open(url);
```
tmcw, line 120: why split these calls?

tristen, line 120: Seemed more legible to me. I will change this though.

tmcw:
> Adds an open option to have attribution turned on by default

Why add this instead of an API to open the attribution control?

> Surfaces the editlink option in infoControl to be on by default

Changes to default behavior are risky and expensive. I don't think this one is entirely justified in light of the risk of confusing people when underlying maps do not use OSM data, which is common.

lxbarth:
> > Adds an open option to have attribution turned on by default
> 
> Why add this instead of an API to open the attribution control?

Yeah not needed to have the info control open by default. Also no API needed to open/close it at this stage.

> when underlying maps do not use OSM data, which is common.

So we just show "Improve this map" when OSM is in the attribution list?

tristen:
> Adds an open option to have attribution turned on by default

Should not have been a part of this pr. I think this feature should be added and will propose it in a separate pr.

tristen:
Voiced this and it makes sense to handle this text from the tilejson layer. Closing in favor of that method.

PR #667, opened 2014-02-04
Examples from support

tristen:
A bunch of examples based on answering some support questions last week I've turned into proper examples.


jfirebaugh:
What the use case for disable-UTFGrid-interaction example? Would it be better to just use:

```
var map = L.mapbox.map('map', 'examples.map-8ced9urs', {gridControl: false});
```

jfirebaugh:
[Toggling UI](https://www.mapbox.com/mapbox.js/example/v1.0.0/toggling-ui/) also shows how to toggle a legend, albeit it has a lot of other features as well.

Is geojson-marker-from-remote-url substantially different than [GeoJSON Marker from URL](https://www.mapbox.com/mapbox.js/example/v1.0.0/geojson-marker-from-url/)?

tmcw:
- Hm, loading from raw.github.com in an example seems like a very bad idea.
- marker-navigation-from-menu-list.html should not use data attributes, it should use variables.
- marker-navigation-from-menu-list.html , on second glance, will not run without errors.

tristen:
Thanks for the review @jfirebaugh 

> What the use case for disable-UTFGrid-interaction example?

The use case would be disabling the default interaction in order to roll with your own thing?

> Toggling UI also shows how to toggle a legend, albeit it has a lot of other features as well.

Totally fine with dropping that one.

> Is geojson-marker-from-remote-url substantially different than GeoJSON Marker from URL?

geojson-marker-from-remote-url demonstrates pulling a geojson file from github which is different from a local resource.

_posts/examples/v1.0.0/0100-01-01-geojson-marker-from-remote-url.html
```diff
@@ -0,0 +1,38 @@
...
+
+<script>
+var url = 'https://api.github.com/repos/mapbox/mapbox.js/contents/test/manual/example.geojson';
+
+var map = L.mapbox.map('map', 'examples.map-h67hf2ic')
+  .setView([0.3941, -78.2227], 7);
+
+function load() {
+  // Fetch just the contents of a .geojson file from GitHub by passing
+  // `application/vnd.github.v3.raw` to the Accept header
```
tmcw, line 22: Add a note about this using CORS and not guaranteed to work in all browsers. It's more important to explain the dumb parts than the complicated parts of examples.

_posts/examples/v1.0.0/0100-01-01-geojson-marker-from-remote-url.html
```diff
@@ -0,0 +1,38 @@
...
+  $.ajax({
+    headers: {
+      'Accept': 'application/vnd.github.v3.raw'
+    },
+    dataType: 'json',
+    url: url,
+    success: function(geojson) {
+        // On success add fetched data to the map.
+        var remoteLayer = L.mapbox.featureLayer().addTo(map);
```
tmcw: Consolidate these two calls: the first argument to `featureLayer` can be a geojson object.

_posts/examples/v1.0.0/0100-01-01-marker-popup-onload.html
```diff
@@ -0,0 +1,33 @@
...
+description: Control whether a marker tooltip should be displayed on load
+tags: [ui, markers]
+---
+
+<div id='map'></div>
+<script>
+var map = L.mapbox.map('map', 'examples.map-h67hf2ic');
+map.setView([40.7271, -73.9840], 14);
+
+// Build a marker from a simple .geojson object:
```
tmcw: should be GeoJSON, not .geojson

PR #671, opened 2014-02-07
Populate attribution control with layer attribution

lxbarth:
Ensure that by default, attribution information is displayed when opting for attributionControl instead of infoControl.

So, in code, when creating a map like so:

```
L.mapbox.map('map', 'examples.c7d2024a', { infoControl: false, attributionControl: true })
```

In the case of the map above, attribution is displayed like so:

screen shot 2014-02-07 at 12 05 24 pm

Right now, when creating map like in the example above, nothing is displayed in the attribution field.

This patch likely needs work as I'm not sure I'm passing the correct information in the call to `addAttribution()`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ BROWSERIFY = node_modules/.bin/browserify
 # the default rule when someone runs simply `make`
 all: \
 	dist/mapbox.js \
-	dist/mapbox.private.js \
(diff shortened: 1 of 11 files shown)


tmcw:
Writing tests brought up some other cruft, so this PR ended up including some other bugfixes and a rename of `private` to `internals`. Alas. Merging.

PR #673, opened 2014-02-08
Improve this map

tristen:
This addition to infoControl listens for the existence of an element `mapbox-improve-map` and attaches the `_editLink` method to it.

### DIFF
--- a/src/info_control.js
+++ b/src/info_control.js
@@ -99,6 +99,9 @@ var InfoControl = L.Control.extend({
     },
 
     _editLink: function() {
+        if (!this._content.getElementsByClassName) {
+            return;
+        }
         var link = this._content.getElementsByClassName('mapbox-improve-map');
         if (link.length && this._map._loaded) {
             var center = this._map.getCenter();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v1.6.2
+
+* Removes `editLink` option for `L.mapbox.infoControl`
+* Recognizes and updates `Improve this Map` link in `L.mapbox.infoControl`
+
 ## v1.6.1
 
 * Adds `select` and `autoselect` events to the geocoderControl to allow for
--- a/test/manual/auto.html
+++ b/test/manual/auto.html
@@ -8,9 +8,9 @@
   <script src="../../dist/mapbox.js"></script>
 </head>
 <body>
-  <div id='map'></div>
-  <script>
-    var map = L.mapbox.map('map', 'examples.map-zr0njcqy');
-  </script>
+<div id='map'></div>
+<script type='text/javascript'>
+    L.mapbox.map('map', 'examples.map-zr0njcqy');
+</script>
 </body>
 </html>
--- a/test/manual/theme.html
+++ b/test/manual/theme.html
@@ -3,7 +3,7 @@
 <head>
 <meta charset=utf-8 />
 <title>Theme | mapbox.js</title>
-<script src='../../dist/mapbox.uncompressed.js'></script>
+<script src='../../dist/mapbox.js'></script>
 <link rel='stylesheet' href='../../dist/mapbox.css'/>
 <style>
   body {margin:0; padding:0; }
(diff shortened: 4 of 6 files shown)


tmcw:
Plz test this with multiple maps, it will break.

tmcw:
Holding on this until we have the server side pushed. And then I'm not really clear on why we're doing the event binding thingy when we could just rewrite the link url.

lxbarth:
> And then I'm not really clear on why we're doing the event binding thingy when we could just rewrite the link url.

I don't care much either way, but @tristen can you sort this out not waiting for server side merge?

tristen:
Sorry, I need more of an explanation here. I shouldn't be listening for the click event to re-write the url with the maps current coordinates?

tmcw:
These tests need to cover when the map actually moves and the code rewrites the link - right now the whole `moveend` event binding is untested.

tmcw:
(that should also cover moveend events with improvemap vs moveend events without the link)

tmcw:
This binds an event `moveend` on the map object but does not unbind it when the control is removed.

tmcw:
It also re-binds that event every time that the map is updated. Let's clean up the event binding situation here.

tristen:
@tmcw I think I'm in a good place here and addressed your comments above.

src/info_control.js
```diff
@@ -43,6 +43,7 @@ var InfoControl = L.Control.extend({
 
     onRemove: function(map) {
         map
+            .off('moveend', this._editLink, this)
```
tmcw, line 46: .off() doesn't have a third argument here

src/info_control.js
```diff
@@ -91,26 +91,24 @@ var InfoControl = L.Control.extend({
...
-        '&lat=' + center.lat + '&lon=' + center.lng);
+    _editLink: function() {
+        var link = this._content.getElementsByClassName('mapbox-improve-map');
+        if (link.length && this._map._loaded) {
+            var center = this._map.getCenter();
+            var tilejson = this._tilejson || this._map._tilejson || {};
+            var id = tilejson.id || '';
+
+            for (var i = 0; i < link.length; i++) {
+                link[i].href = 'https://www.mapbox.com/map-feedback/#' + id + '/' + center.lng + '/' + center.lat + '/' + this._map.getZoom();
```
lxbarth: @tristen - any reason we echo the URL here? Let's just add to the URL instead of replacing it, no?

lxbarth:
@tmcw - is this good to merge?

src/info_control.js
```diff
@@ -91,26 +91,25 @@ var InfoControl = L.Control.extend({
...
-        '&lat=' + center.lat + '&lon=' + center.lng);
+    _editLink: function() {
+        var link = this._content.getElementsByClassName('mapbox-improve-map');
+        if (link.length && this._map._loaded) {
+            var center = this._map.getCenter();
+            var tilejson = this._tilejson || this._map._tilejson || {};
+            var id = tilejson.id || '';
+
+            for (var i = 0; i < link.length; i++) {
+                if (!this._editurl) this._editurl = link[i].href;
```
tmcw: Can you make this less brittle?

``` js
link[i].href = link[i].href.split('#') + '#' + id + '/' + center.lng + '/' + center.lat + '/' + this._map.getZoom();
```

tmcw:
@lxbarth i haven't had the time to test it along with the tilestream-pro patch, and ideally that patch is merged first. The code looks okay except for ^ where it could be improved, but I'd need to see it actually working.

lxbarth:
Awesome.

@tristen - please [address brittleness](https://github.com/mapbox/mapbox.js/pull/673#discussion_r10013638). This PR should be ready to be merged then.

@tmcw let's do voice on roll out order.

src/info_control.js
mourner, line 46: Why not? I think there should be a third argument on all `off` calls if it was given in corresponding `on`.

lxbarth:
@tmcw - after voice just now:
- Removed any irrelevant changes (mostly whitespace and debug stuff)
- Switched test to local server
- Simplified tests
- Tested whether infoControl works on old and new attribution (manually and in unit tests)
- Protected call to `getElementsByClassName` so IE8 or other older browsers don't choke
- Spot checked IE8 on browserstack

@tmcw - please review.

/cc @tristen

PR #678, opened 2014-02-12
Docs navigation

tristen:
Revising of the docs header:
- Re-introduce the mapbox masthead on docs pages.
- Spruce the overview page on /api/ by condensing the frontpage into it.
- Play off the green headline for site title.

PR #681, opened 2014-02-18
Docs navigation

tristen:
This is a scaled back commit as reflect in comments [here](https://github.com/mapbox/mapbox.js/pull/678)
Improved navigation flow with fixed navigation as you scroll down the page.
### API

screen shot 2014-02-18 at 10 55 54 am
### Examples

screen shot 2014-02-18 at 10 56 38 am
### Plugins

screen shot 2014-02-18 at 10 56 50 am

cc/ @lxbarth @tmcw


lxbarth:
@tristen - there are merge conflicts. Can you resolve?

lxbarth:
Great, this feels close. 
- When scrolling down, width of sidebar changes.
- Landing page could benefit from a lot more space, just push things down the page and give the intro more breathing room. 
- Can we get away without the fixed sidebar toggle? You're right, it feels awkward. Check out this animated gif where the fixed toggle is removed and the sidebar is just scrollable. I have a hunch tho I'm missing a finer point in CSS :)

scroll

tristen:
> When scrolling down, width of sidebar changes.

Our col widths are percentage based. The side bar is 25% of the `.limiter` class width (83.3333%) When the sidebar element changes to a fixed position it's 25% is relative to the screen width. In order to counteract this I need to apply a new percentage width that is a fixed 25% of 83.3333% (0.20833325) the hairline change in width is this 'accurate as possible' adjustment to the previous width.

The only real way around this would be to switch over to pixel units and not use percentages in our doc columns. That's a shame.

> Landing page could benefit from a lot more space, just push things down the page and give the intro more breathing room.

The intro content has 40px of bottom space between it and the **Reading the Documentation** section. This is consistent pixel spacing between sections across all our sites. Let's not break this with some weird extra spacing.

> Can we get away without the fixed sidebar toggle? You're right, it feels awkward. Check out this animated gif where the fixed toggle is removed and the sidebar is just scrollable. I have a hunch tho I'm missing a finer point in CSS :)

What would that look like? The sidebar would remain fixed on the page:

fixed-layout

tmcw:
Can't we add a little bit of js magic to reposition the sidebar to the top when the page scrolls?

yhahn:
@tristen @lxbarth @tmcw maybe I am missing a lot of context but nothing about this seems shippable. Every aspect of the design seems compromised by some other aspect -- the branding, the site nav, the api nav, the first impression.

I _will_ jump off whatever I am working on to help here if needed.

tristen:
Let's take a step back. This pull request is not getting us to where we need to be and is a hotfix way of solving one problem: long navigation.  

There's an opportunity here to create great documentation that nails first impression and feels cohesive to the entire experience of creating and building with mapbox. I'm  interested in proposing something different and will sketch out ideas that communicate this intent stronger.

tristen:
Move to #692

PR #718, opened 2014-03-19
Don't use _createPane for arbitrary element creation

jfirebaugh:
Best not to use `_` methods -- `_createPane` is gone in Leaflet master. (There's a public `createPane` method but it does something slightly different.)

### DIFF
--- a/src/share_control.js
+++ b/src/share_control.js
@@ -27,9 +27,9 @@ var ShareControl = L.Control.extend({
         var link = L.DomUtil.create('a', 'mapbox-share mapbox-icon mapbox-icon-share', container);
         link.href = '#';
 
-        this._modal = map._createPane('mapbox-modal', this._map._container);
-        this._mask = map._createPane('mapbox-modal-mask', this._modal);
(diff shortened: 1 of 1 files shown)


mourner:
:+1:

jfirebaugh:
@tristen Anything jogging your memory here? Seems to work fine for me.

tristen:
Just going to test this real quick. If everything seems fine this should be good2go.

tristen:
Yep - this looks great to me!

wboykinm:
With this in mind, what do you recommend I use [in place of `_createPane` here?](https://github.com/wboykinm/reference/blob/master/mapbox.html#L174)

jfirebaugh:
Just inline the definition of [_createPane](https://github.com/Leaflet/Leaflet/blob/6864384d630e46fc5908575befebb19239b1c0d0/src/map/Map.js#L525), so:

```
var topPane = L.DomUtil.create('div', 'leaflet-top-pane', map.getPanes().mapPane);
```

PR #725, opened 2014-03-31
Export mapbox.js for browserify

tmcw:


schickling:
Would be such a powerful thing...

tmcw:
In case you're wondering, this doesn't fix the issue. Leaflet window-exports itself, and `mapbox.js` does not declare a module.exports.

Issue #744, opened 2014-05-12
Expanded attribution by default


tristen:
via http://www.openstreetmap.org/user/lxbarth/diary/21847
- [x] Make attribution control default instead of info control
- [x] Get **Improve this map** link also working on attribution control

If this is always present, I also want to tone the color from the blue to a 'read' color value with text-decoration:underline on hover.

cc/ @tmcw @lxbarth

tristen:
https://github.com/mapbox/mapbox.js/pull/748

PR #748, opened 2014-05-13
Revised attribution

tristen:
Hits https://github.com/mapbox/mapbox.js/issues/744
- Attribution control is now default over of infoControl.
- infoControl is now triggered when it's set to true in the `L.mapbox.map` options param or when passed as an argument to `map.addControl()`.
- **Improve this map** link now tracks on the attribution control
### TODO
- [x] Mark change in CHANGELOG
- [x] Test coverage for **Improve this map** in L.control.attribution.

cc/ @tmcw for a review.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -132,7 +132,7 @@ var LMap = L.Map.extend({
         var link = this._controlContainer.getElementsByClassName('mapbox-improve-map');
         if (link.length && this._loaded) {
             var center = this.getCenter().wrap();
-            var tilejson = this._tilejson || this._map._tilejson || {};
+            var tilejson = this._tilejson || {};
             var id = tilejson.id || '';
 
             for (var i = 0; i < link.length; i++) {
@@ -155,7 +155,6 @@ var LMap = L.Map.extend({
             this.attributionControl.addAttribution(layer.getAttribution());
         }
 
-        this.off('moveend', this._editLink, this);
         this.on('moveend', this._editLink, this);
 
         if (!(L.stamp(layer) in this._zoomBoundLayers) &&
--- a/src/map.js
+++ b/src/map.js
@@ -128,9 +128,7 @@ var LMap = L.Map.extend({
     },
 
     _editLink: function() {
-        if (!this._controlContainer.getElementsByClassName) {
-            return;
-        }
+        if (!this._controlContainer.getElementsByClassName) return;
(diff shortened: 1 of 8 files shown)


src/map.js
```diff
@@ -139,11 +155,15 @@ var LMap = L.Map.extend({
             this.attributionControl.addAttribution(layer.getAttribution());
         }
 
+        this.off('moveend', this._editLink, this);
```
tmcw: What is this doing - why unbind and rebind the same function to the same event?

test/spec/info_control.js
```diff
@@ -93,7 +95,9 @@ describe('L.mapbox.infoControl', function() {
     it('adds map location to improve link in attribution when .mapbox-improve-map is present', function(done) {
         var server = sinon.fakeServer.create();
         var element = document.createElement('div');
-        var map = L.mapbox.map(element, 'examples.h8e9h88l');
```
tmcw, line 96: Doesn't look like any test coverage is added for the ability to click 'improve this map' in L.control.attribution, not just infoControl

tmcw:
Made inline comments for potentially tricky sections of code. Besides that, this'll need an update in the CHANGELOG.md.

src/map.js
```diff
@@ -128,6 +127,23 @@ var LMap = L.Map.extend({
         }
     },
 
+    _editLink: function() {
+        if (!this._controlContainer.getElementsByClassName) return;
+        var link = this._controlContainer.getElementsByClassName('mapbox-improve-map');
+        if (link.length && this._loaded) {
+            var center = this.getCenter().wrap();
+            var tilejson = this._tilejson || this._map._tilejson || {};
```
tmcw: `this._map._tilejson` will never mean anything here.

tmcw:
Okay: my only remaining concern is that the visual difference here will be quite significant for long attribution strings:

2014-05-14 at 2 03 pm

These are, of course, the minority, and might be fine as-is. Thoughts @tristen + @lxbarth ?

tristen:
That seems like a infoControl instead of L.control.attribution use case.

tmcw:
Yes, but the concern is that by switching these on Mapbox.com, we'll turn an (i) into the above - there's no plan of making this a user-facing UI option (do you want less attribution or more?), so should we be fine with long long strings, or should we do something more auto?

tristen:
My heart is telling me to merge for now. The auto solution could be good if this turns into a bad problem.

Issue #695, opened 2014-03-01
Chrome popping up warning for using dpi instead of dppx in CSS


antidis:
Chrome is having some trouble with the Mapbox CSS. In Chrome 33.0.1750.117/Mac, and Chrome 35.0.1866.0 canary/Mac, the following warning pops up as a result of the Mapbox CSS being loaded:

"Consider using 'dppx' units instead of 'dpi', as in CSS 'dpi' means dots-per-CSS-inch, not dots-per-physical-inch, so does not correspond to the actual 'dpi' of a screen. In media query expression: (-webkit-min-device-pixel-ratio: 2), (min-resolution: 192dpi)"

The source of this line is here:

``` css
@media
  (-webkit-min-device-pixel-ratio:2),
  (min-resolution:192dpi)
```

...which is missing a dppx declaration. The equivalent "future-proof" query, which _should_ work, is below:

``` css
@media
(-webkit-min-device-pixel-ratio: 2),
(min--moz-device-pixel-ratio: 2),
(-o-min-device-pixel-ratio: 2/1),
(min-device-pixel-ratio: 2),
(min-resolution: 192dpi),
(min-resolution: 2dppx)
```

...as detailed here: http://css-tricks.com/snippets/css/retina-display-media-query/

I say _should_ because Chrome seems to be falling over anyway, acting as if dppx isn't referenced: https://code.google.com/p/chromium/issues/detail?id=336276

Would it be possible to update the CSS to avoid this warning, once Chrome fixes its warning bug?

tmcw:
@tristen does your commit above fix this?

tristen:
I just fixed up that commit preserving 192dpi and adding 2dppx. Will return to this and make any adjustments when the warning bug is addressed.

antidis:
Thanks @tristen

awallace-ihouse:
So this hasn't made it into the production code yet? I see it on my pages now, using:
    https://api.tiles.mapbox.com/mapbox.js/v1.6.2/mapbox.css

antidis:
It has. Chrome is still reporting this error, but incorrectly so. Chrome will need to fix this on their end too, but it looks like there's a fix in review: https://code.google.com/p/chromium/issues/detail?id=336276

Chrome also reports it on its own default tab screen (!!!)

awallace-ihouse:
Thanks for the info! Luckily it’s only annoying, as opposed to fatal.

————————————
Andrew Wallace
awallace@cisdata.net

[...]

PR #775, opened 2014-05-23
Drop retina conditions and just serve a @2x version

tristen:
Tested and I'm not seeing any disadvantage in doing this. Addresses https://github.com/mapbox/mapbox.js/issues/695 too.

cc @tmcw


tmcw:
Hmm, I'm not sure that scaling down will really always look exactly the same... thoughts @samanpwbb ?

samanpwbb:
@tmcw scaling down looks exactly the same as long as 2x version is even length/width. Browser rendering fails when the 1/2 size version doesn't fall on the pixel grid. In this case, it should look right.

tmcw:
Cool, then merge it @tristen

tmcw:
This looks like it broke http://localhost:3000/test/manual/standalone.html -

2014-05-27 at 3 52 pm

Issue #766, opened 2014-05-20
detectRetina: true by default


tmcw:
for 1.6.4: we should finally make the jump here. Sharp by default.

tmcw:
Haha, so, uh,

|  | detectRetina=false | detectRetina=true | detectRetina=undefined |
| --- | --- | --- | --- |
| retinaVersion=true | no | yes | yes |
| retinaVersion=false | no | yes | no |

tmcw:
Fixed in #781

PR #781, opened 2014-05-27
Try to make autoscale=true a default

tmcw:
This ends up being more of a quagmire than expected - we initialize
tileLayers in L.mapbox.map and L.mapbox.tileLayer, while Leaflet expects
us to know retina status right then, and doesn't seem to support
changing this value after it is set. Would love some ideas from
@mourner about how this could become less hacky.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "main": "index.js",
   "dependencies": {
-    "leaflet": "0.7.2",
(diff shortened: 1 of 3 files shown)


tmcw:
This aims to fix #766 - when TileJSON says autoscale is available and the user has not explicitly said no retina, we should use retina tiles.

PR #800, opened 2014-06-13
Use zuul for cross-browser testing.

tmcw:
This doesn't include IE yet, because sinon-ie will break non-ies if included always


tmcw:
Sauce Labs sucks.

PR #826, opened 2014-07-21
Document using mapbox.js from the CDN

eschwartz:
And describe difference between mapbox.js and mapbox.standalone.js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,18 @@ Browserify it:
 browserify main.js -o bundle.js
 ```
 
+## Usage from our CDN
+
+Load mapbox.js from our CDN:
+
+```html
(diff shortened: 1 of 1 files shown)


eschwartz:
I'm using MapBox for a RequireJS project. MapBox is an optional dependency (some builds include it, others do not), so I needed to be be able to load Leaflet independently of MapBox.

I ran into the `mapbox.standalone.js` build by accident, which fixed my problems. I'm hoping this update will help anyone else with the same issue.

jfirebaugh:
Thanks!

tmcw:
Can you do a cleanup for the grammar and style issues in this PR?

jfirebaugh:
I did a followup commit.

Issue #740, opened 2014-05-02
Provide Compiled Version of code for Bower


willfarrell:
Currently I'm using `https://github.com/platanus/bower-mapboxjs` for bower since I don't use requirejs. Is it possible to update your build process to generate `/dist/*.{js,css}` files? I prefer to get the code from the creators. Thanks again for an awesome service.

tmcw:
See https://github.com/mapbox/mapbox.js/issues/404 - happy to accept a pull request that does this properly, but otherwise you can use bower install with http urls.

willfarrell:
Thanks, I'll see what @platanus is doing. Maybe they can create a PR.

Issue #404, opened 2013-06-01
Add to Bower


tysonnero:
It would be great to get mapbox.js into the Bower registry :-)

tmcw:
We've discussed this a bit, and are open to the possibility but it's not as easy as adding a `bower.json` - right now mapbox.js does not include dists in its repo and is built with browserify. This will have to change if we support bower, and our tagging strategy will have to get more sophisticated as well.

tmcw:
Bower is more trouble than it's worth. Closing.

adamnbowen:
For anyone interested in cheating this into your bower.json file without requiring browserify, you can run this in the command line:

```
$ bower install mapbox.js=https://api.tiles.mapbox.com/mapbox.js/v1.4.2/mapbox.js --save
```

and

```
$ bower install mapbox.css=https://api.tiles.mapbox.com/mapbox.js/v1.4.2/mapbox.css --save
```

That will let you incorporate the compiled files into your usual build process.

That doesn't get you the required images, however, so I'll have to investigate that a bit.

amir20:
I think this ticket should be reopened because I don't think it is more trouble that one thinks. You could do what angularjs does which have a distribution just for the compiled version. Seems like you can automate it for every release.

tmcw:
I'm open to a PR that implements & automates this process, or a third-party adapter.

Anahkiasen:
You could just add the ZIP of each release to the Github releases in the meantime, that would make them easier to find.

nerik:
Here's a quick and dirty grunt task to make a local copy of the mapbox assets (including images)
https://gist.github.com/nerik/64e4798d2bf1c19800f6
(also does other things related to our assets pipeline)

sanketsahu:
Bower support is much needed.

Nolski:
+1 for bower support. I understand that it may take time due to the way things are set up but closing a ticket because it takes work doesn't seem like a valid excuse to me.

tmcw:
`¯\_(ツ)_/¯` I'll reopen the ticket, but as before, open to PRs here - they get stuff done faster than +1ing

PR #830, opened 2014-07-25
Adds build script to create and publish the bower repo

blackjid:
Hi, we at platanus have been running the `platanus/bower-mapboxjs` repo that is being used by the `mapboxjs` package from bower. Sadly we're not using mapbox to actively so the repo usually gets outdated.

I've build a script that automate the bower building and publishing trying to solve this issues platanus/bower-mapboxjs#2 and #404 and #740

If you are willing to merge the PR you can fork our repo, or if you prefer I can transfer the ownership to you.

Also we can change the endpoint to `mapboxjs` in bower

I hope this is useful for everyone

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 node_modules
 _site
+tmp
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -20,3 +20,8 @@ Then make the release with `make`
 # deploying to the cdn
 
     ./deploy.sh v1.your.version
+
+# deploying to bower
+
+    ./bower.sh --action=prepare
+    ./bower.sh --action=publish
--- a/bower-utils.inc
+++ b/bower-utils.inc
@@ -1,0 +1,279 @@
+# This file provides:
+# - a default control flow
+#   * initializes the environment
+#   * able to mock "git push" in your script and in all sub scripts
+#   * call a function in your script based on the arguments
+# - named argument parsing and automatic generation of the "usage" for your script
+# - intercepting "git push" in your script and all sub scripts
+# - utility functions
+#
+# Usage:
+# - define the variable ARGS_DEF (see below) with the arguments for your script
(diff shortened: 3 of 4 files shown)


tmcw:
Awesome - I will review this next week and scope out merging / running it on an npm publish task

blackjid:
Great, let me know if you need anything.

tmcw:
Okay, took a look:
- This PR creates a `master` branch in which bower releases live. We don't currently have a `master` branch, and I think it would be confusing for `master` to be bower-only. I'd love to use a branch like `bower` that's specific about being bower-specific so that nobody ends up accidentally cloning and developing on `master`.
- The script also creates new git tags of the [same format as we do already](https://github.com/mapbox/mapbox.js/releases) - I know bower is git-tag-dependent, and thus would lead us into the paradox of which branch the git tag should live on. Maybe bower releases could be like `2.0.1-bower`, which is 'pre-release' according to [semver](http://semver.org/), but doesn't sacrifice the existing patterns.

tmcw:
Ah, now I notice, too late, that this is checking out & pushing to a separate repo - got it.

tmcw:
@blackjid thanks, merged & set up at https://github.com/mapbox/mapbox.js-bower - can you [unregister the existing distro so we can adopt it?](http://bower.io/docs/creating-packages/#unregister)

blackjid:
:+1: 

If you want to take over the ownership of [platanus/bower-mapboxjs](https://github.com/platanus/bower-mapboxjs) just let me know. There I have build all mapbox.js tags with this script so bower users can point to any version.

tmcw:
Good call - forked that repo to https://github.com/mapbox/mapbox.js-bower

blackjid:
done. bower `mapboxjs` no longer taken..

mrpollo:
Great work guys thanks, this works great :sob:

ap

Issue #798, opened 2014-06-12
broken link on intro page for previous versions


jfirebaugh:
E.g. on https://www.mapbox.com/mapbox.js/api/v1.6.3/, "Check out v1.6.4 for the latest" links to https://www.mapbox.com/mapbox.js/api/v1.6.4/v1.6.3/.

PR #831, opened 2014-07-28
removes logic that was borking links for older versions

geografa:


jfirebaugh:
Refs #798. Problem is this logic _is_ needed for sub-pages like https://www.mapbox.com/mapbox.js/api/v1.6.3/l-mapbox-tilelayer/

geografa:
aha. k, I'll close this up since it's on the radar.

Issue #851, opened 2014-08-18
Geocoder results not populating result name in Firefox


tristen:
![screen shot 2014-08-18 at 10 07 54 am](https://cloud.githubusercontent.com/assets/61150/3952353/4eac1944-26e1-11e4-82d7-08647085cd16.png)

PR #853, opened 2014-08-18
Use innerHTML instead of innerText

tristen:
Address' https://github.com/mapbox/mapbox.js/issues/851

### DIFF
--- a/src/geocoder_control.js
+++ b/src/geocoder_control.js
@@ -112,7 +112,8 @@ var GeocoderControl = L.Control.extend({
                         if (!name.length) continue;
 
                         var r = L.DomUtil.create('a', '', this._results);
-                        r.innerText = name;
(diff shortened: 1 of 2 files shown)


tmcw:
Hm. Let's fallback to `.textContent` instead: using innerHTML will change behavior here and make this a potential security hole.

jfirebaugh:
Pretty sure we should just use `textContent` unconditionally. `innerText` was my bad.

tmcw:
innerText would give us IE8 support here - seems like it's worth it. IE8+ is still the advertised compatibility range.

jfirebaugh:
Oh, I missed that. :+1:, carry on.

jfirebaugh:
Hmm, diff now looks wrong though. `grid_control.js`?

PR #869, opened 2014-09-04
Allow map to absorb options from mergeOptions. Fixes #829

### LINKED ISSUE #829: L.mapbox.Map does not use L.Map options after initialization
When I attempt to merge options in L.Map and then create a map instance, the options are not present on the instance.

For example:
- Load mapbox.js
- Call: L.Map.mergeOptions({ test: true });
- Init map: L.mapbox.map(element, defaultLayerId, mapOptions);

this.options.test -> does NOT exist
L.Map.prototype.options.test -> exists
L.mapbox.Map.prototype.options.test -> does NOT exist

I understand why this happens. When L.Map is extended, the prototype options are copied over but any further updates to the prototype options will be ignored by L.mapbox.Map.

I also know I can merge the options to L.mapbox.Map, however I don't want to start modifying 3rd party leaflet plugins to make sure the default options work.

tmcw:


jfirebaugh:
FWIW, this will be fixed upstream in the next major Leaflet release. https://github.com/Leaflet/Leaflet/commit/ad9d0f8c7fbb080fa628f302bddb52e6c035c9d2

PR #870, opened 2014-09-05
Update bower.json

teawithfruit:


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mapbox.js",
-  "main": "index.js",
+  "main": "mapbox.js",
```
jfirebaugh, line 3: Shouldn't it be `"src/index.js"`? There is no `mapbox.js` file at the root level.

tmcw, line 3: Our bower rubegoldbergmachine publishes releases over at [this repo](https://github.com/mapbox/mapbox.js-bower), where there is a root-level mapbox.js

jfirebaugh:
Ok, thanks!

PR #871, opened 2014-09-08
First shot at shooting stuff over to jsfiddle

tmcw:
- [x] remove or restyle control when user isn't logged in
- [x] clean up and separate html & js so that jsfiddle doesn't complain


tmcw:
rebased and merged in b1ad299087df2111111e5368154dfa31dda7e70b

PR #886, opened 2014-10-06
Remove examples/redirects/v0.6.7

tristen:
These old title's show up in mapbox.com/help search and rather than fiddling around with limiting them in swiftype I think we should just remove them from the project.


jfirebaugh:
:+1:

PR #887, opened 2014-10-06
Event listener example for geocoder control

tristen:
This example demonstrates using the `found` event from Geocoder control.


tmcw:
:+1: cool, go for it.

PR #889, opened 2014-10-08
fix typo in GeoJSON static functions

whyvez:
Found these typos while working with mapbox.js. Hopefully this is the right branch to PR to :)


jfirebaugh:
Thanks @whyvez. This is a copy of [a file from the Leaflet repo](https://github.com/Leaflet/Leaflet/blob/gh-pages/reference.html), so the thing to do is 1) fix it there and 2) update our copy. In fact it's [already been fixed upstream](https://github.com/Leaflet/Leaflet/commit/286d06dbc3ba1b77c836919173ea443c70e9f94f), so we just need to do the second step.

whyvez:
OK perfect.

PR #891, opened 2014-10-09
add 2x zoom control for small screen

whyvez:
Added styles to double zoom tap targets for smaller screens. I have fat fingers... ;) I'm sure it would help people with normal size fingers too. Was requested on a project I'm working on and thought I might as well do the fix here. Please ignore if deemed not valuable.

### DIFF
--- a/theme/style.css
+++ b/theme/style.css
@@ -758,6 +758,15 @@
   height:100%;
   margin:0;
   }
+.leaflet-control-zoom.leaflet-bar a {
+  height: 52px;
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Hi @whyvez, thanks for the contribution. I think this is something best handled on a per-project basis with CSS overrides. If we were going to consider larger tap targets for Mapbox.js core it would need to be a unified approach across other controls in addition to the zoom controls.

PR #894, opened 2014-10-10
Trigger the `found` event when autocomplete is enabled.

tristen:


tmcw:
`autoselect` means that you searched for a word, found on thing, and the geocoder automatically brought you there. when you have autocomplete enabled, the geocoder does not automatically go to the first result.

tristen:
Oh dur

PR #896, opened 2014-10-16
Include mapbox.css in the main entry

derekcicerone-zz:
This allows tools like wiredep to automatically inject all the mapbox resources into a web page.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,9 @@
 {
   "name": "mapbox.js",
-  "main": "mapbox.js",
+  "main": [
+    "mapbox.css",
+    "mapbox.js"
+  ],
   "version": "",
   "homepage": "https://github.com/mapbox/mapbox.js",
   "authors": ["Mapbox"],


tmcw:
Hm: https://www.npmjs.org/doc/files/package.json.html#main - it seems very likely that wiredep is simply doing it wrong.

derekcicerone-zz:
That's the NPM documentation. Bower is inspired by NPM syntax for package.json but the main entry works differently.

For example, here is leaflet's bower.json: https://github.com/Leaflet/Leaflet/blob/master/build/bower.json

derekcicerone-zz:
Here is the bower documentation for this:

```
main string or array: The primary acting files necessary to use your package.
```

(from http://bower.io/docs/creating-packages/#bowerjson)

tmcw:
Oh, it's for bower. All right then.

derekcicerone-zz:
:) thanks!

PR #898, opened 2014-10-20
Condition IE8 to use base64 encoded sprites

### LINKED ISSUE #897: Zoom Control not styled correctly in IE8
jira-capture-screenshot-20141015-163213-687

tristen:
Closes #897 by [conditioning](http://browserhacks.com/#hack-c9242a5ec3f073257e275102be15d95f) IE8 to use base64 encoded sprites.


jfirebaugh:
@tristen When adding browser-specific hacks, please record the background/rationale in the pull-request  or commit comment. I.e. what exactly the issue is, what conditions trigger it, what less-hacky solutions were tried and why they didn't work. Eventually we want to retire support for legacy browsers and these kinds of hacks and recording that information will make it that easier to do in the future.

tristen:
@jfirebaugh :+1:  I should have justified this. 

IE < 8 doesn't support the `background-image` property which is used to scale the 180dpi exported sprite for retina screens. As this is the only browser (we support) that doesn't support this property I went with [our convention](https://github.com/mapbox/mapbox.js/blob/8613549cabeb5228b6ca76dbc9ea73e0205f02d0/theme/style.css#L763-L799) of adding the browser hack to the bottom of mapbox.css. a base64 encoded sprite seemed like the best way to go as maintaining another asset would suck.

The more traditional CSS hack ie. `hack { width:20px\9; }` didn't work as expected for `background-image` so I wrapped this in a media query hack.

PR #899, opened 2014-10-21
v2.2.0

tristen:
@tmcw @jfirebaugh :+1: :-1: ?


jfirebaugh:
The `found` event change feels more like a bugfix than a new feature to me, so in terms of semver this should probably be 2.1.3. WDYT?

tristen:
Yeah, agreed.

PR #909, opened 2014-11-10
listen for access token dom changes, then copy

bsudekum:
This change will postpone getting html/js from the example until the access token has been updated.

Works for both auth'd and non-auth'd users alike.

Ref: https://github.com/mapbox/mapbox.js/issues/847

cc @jfirebaugh

### DIFF
--- a/docs/assets/js/site.js
+++ b/docs/assets/js/site.js
@@ -17,23 +17,25 @@ function load() {
         forceHandCursor: true
     });
 
-    $('.js-clipboard').each(function() {
-        var $clip = $(this);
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I don't think this is a good approach. Mutation events like `DOMNodeInserted` are [deprecated and could be removed](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events).

Looks like we can use [`data-clipboard-target`](https://github.com/zeroclipboard/zeroclipboard/blob/master/docs/instructions.md#text-to-copy) and have the text pulled directly from the DOM when the button is clicked. ⇢ #910

PR #910, opened 2014-11-10
Use data-clipboard-target to pull pull code with access token (fixes #847)

### LINKED ISSUE #847: copy example button should copy example with a token in it

jfirebaugh:


tristen:
Niiiccee

PR #922, opened 2014-12-15
Fixed syntax error in css stylesheet

runarberg:
When compiling with LESS I got a syntax error.


tristen:
LGTM. This makes things more obvious with the className.

runarberg:
LESS is a proper superset of valid CSS. LESS rightly recognizes this as a syntax error and doesn't compile this, even though old IE browsers wrongly parse this.

jfirebaugh:
I like the idea of getting rid of the selector hack. Let's use the already-existing [`.leaflet-oldie`](https://github.com/Leaflet/Leaflet/blob/72feb7005dece9665a7a9292d5a17fde32c3d923/src/map/Map.js#L454) class instead of introducing `.leaflet-IE8down` though.

src/map.js
```diff
@@ -34,7 +34,7 @@ var LMap = L.Map.extend({
 
         L.Map.prototype.initialize.call(this, element,
             L.extend({}, L.Map.prototype.options, options));
-
+
```
tmcw, line 37: Can you exclude this file from the commit? This is just a whitespace change.

jfirebaugh:
Cleaned up and committed as 1e69a8d554f87da9556633bc8a9c494b4efa81f3. Thank you!

PR #925, opened 2014-12-20
Fixes error on local usage of mapbox css - Makes img link in CSS an explicit url instead of relative

drewrwilson:
Right now if you use mapboxjs and mapboxcss locally instead of through the CDN it works well, but it gives one error trying to load this img: `/images/icons-000000@2x.png`. It would require anyone using mapbox css locally to add a directory called `images/` in their css directory and then download that icon set and put it in there. So by making it the link to the icon an explicit link instead of relative, we can use mapbox css locally without also needing to download that file, the css will just get the icon from your tile server.

### DIFF
--- a/theme/style.css
+++ b/theme/style.css
@@ -399,7 +399,7 @@
 .map-tooltip .close,
 .mapbox-icon {
   opacity: .75;
-  background-image:url(./images/icons-000000@2x.png);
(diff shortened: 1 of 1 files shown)


tmcw:
People should download `/images/` if they want to run mapbox.js locally: having two files local and one remote doesn't make sense for firewalled or offline usecases any more than having three remote files does.

PR #931, opened 2015-01-13
Watermark

camilleanne:


src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
...
+
+        this._container.innerHTML = this._mapboxLogo;
+
+        return this;
+    }
+});
+
+module.exports.MapboxLogo = MapboxLogo;
+
+module.exports.MapboxLogo = function(options) {
```
jfirebaugh: `module.exports.mapboxLogo = function(options)` (initial lowercase for factory functions).

src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+var MapboxLogo = L.Control.extend({
+
+    options: {
+        position: 'bottomleft',
+        sanitizer: require('sanitize-caja')
```
jfirebaugh: The `sanitizer` option isn't used and can be removed.

src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+var MapboxLogo = L.Control.extend({
```
tmcw: Can this be called MapboxLogoControl to be more in line with the rest and clearly indicate that it's a control?

src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
...
+
+    options: {
+        position: 'bottomleft',
+        sanitizer: require('sanitize-caja')
+    },
+
+    initialize: function(_) {
+        this._mapboxLogo = '';
+        this._ = _;
```
tmcw: what is _? can this be renamed options? Also, note that the L.Control class has a `.setOptions` method that this should probably use instead

test/spec/mapbox_logo.js
```diff
@@ -0,0 +1,83 @@
+describe('mapbox_logo', function() {
```
jfirebaugh, line 1: To get this file to be picked up by the test runner, add `<script src="spec/mapbox_logo.js"></script>` to `test/index.html`.

test/spec/mapbox_logo.js
```diff
@@ -0,0 +1,83 @@
...
+
+    afterEach(function() {
+        server.restore();
+    });
+
+    it('constructor', function() {
+        var map = L.mapbox.map(element, tileJSON);
+        var mapboxLogoControl = map._mapboxLogoControl;
+        expect(map._mapboxLogoControl instanceof L.Control);
+        expect(map.getTileJSON()).to.eql(helpers.tileJSON);
```
jfirebaugh: Remove the assertions about `map.getTileJSON()` -- the behavior of `getTileJSON()` is not pertinent to MapboxLogoControl tests.

test/spec/mapbox_logo.js
```diff
@@ -0,0 +1,83 @@
...
+        var map = L.mapbox.map(element, tileJSON);
+        var mapboxLogoControl = map._mapboxLogoControl;
+        expect(mapboxLogoControl._mapboxLogo.length).to.eql(0);
+        expect(map.getTileJSON()).to.eql(helpers.tileJSON);
+    });
+
+    it('is on tilejson map with mapbox_logo === true', function() {
+        var map = L.mapbox.map(element, helpers.tileJSON_mapboxlogo);
+        var mapboxLogoControl = map._mapboxLogoControl;
+        expect(mapboxLogoControl._mapboxLogo.length).to.eql(6396);
```
jfirebaugh: Change this to assert `length > 0` or something else that won't start failing if the logo gets visually tweaked.

src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
...
+
+        this._update();
+
+        return this._container;
+    },
+
+    addLogo: function(json) {
+        // If account referenced by the accessToken is
+        // asscociated with the Mapbox Logo, display logo.
+        var mapid = typeof this.options.urlOrMapID === 'string' && this.options.urlOrMapID.indexOf('/') === -1 ? true : false;
```
jfirebaugh: Don't need the `? true : false`, here: `typeof this.options.urlOrMapID === 'string' && this.options.urlOrMapID.indexOf('/') === -1` is already a boolean value.

src/mapbox_logo.js
```diff
@@ -0,0 +1,46 @@
...
...LjUgMS45NyA0LjUzIDQuMzggNC41MyAyLjQxIDAgNC4zNCAtMi4wMyA0LjM0IC00LjUzIDAgLTIuNSAtMS45NCAtNC41MyAtNC4zNCAtNC41MyB6IG0gLTIyIDAuMDMgMCAxMiAxLjc1IDAgMCAtMi45NyBjIDAuNTcgMCAxLjA0IC0wIDEuNTkgMCAyLjQ0IDAgNC4zNCAtMS45MSA0LjM0IC00LjUzIDAgLTIuNjIgLTEuOSAtNC41IC00LjM0IC00LjUgeiBtIDI2LjE2IDAgMy4wMyA0LjM4IC0zLjE5IDQuNjIgMi4xMiAwIEwgNTcuMzEgMTEuOTEgNTkuNDQgMTUgNjEuNTkgMTUgNTguMzggMTAuMzggNjEuNDEgNiA1OS4yNSA2IDU3LjMxIDguODEgNTUuMzQgNiB6IE0gMjEuNDcgNy43MiBjIDEuNCAwIDIuNDQgMS4xOSAyLjQ0IDIuNzggbCAwIDIuNzggLTIuNDQgMCBjIC0xLjQgMCAtMi40NCAtMS4yMSAtMi40NCAtMi43OCAtMCAtMS41NyAxLjA0IC0yLjc4IDIuNDQgLTIuNzggeiBtIDI3LjU2IDAgYyAxLjQ0IDAgMi41OSAxLjI0IDIuNTkgMi43OCAwIDEuNTQgLTEuMTUgMi43OCAtMi41OSAyLjc4IC0xLjQ0IDAgLTIuNjIgLTEuMjQgLTIuNjIgLTIuNzggMCAtMS41NCAxLjE4IC0yLjc4IDIuNjIgLTIuNzggeiBtIC0yMC4yNSAwLjAzIDEuNTkgMCBjIDEuNTkgMCAyLjU5IDEuMjggMi41OSAyLjc1IDAgMS40NyAtMS4xMyAyLjc4IC0yLjU5IDIuNzggbCAtMS41OSAwIHogbSA4Ljk3IDAgMS41IDAgYyAxLjQ3IDAgMi42MiAxLjI4IDIuNjIgMi43NSAwIDEuNDcgLTEuMDQgMi43OCAtMi42MiAyLjc4IGwgLTEuNSAwIHoiIHRyYW5zZm9ybT0ibWF0cml4KDUuNzMzMzQxNCwwLDAsNS43MzMzNDE0LDIzNi45MzMwOCwzOTcuMTc0OTgpIiBzdHlsZT0iZmlsbDojZmZmZmZmO2ZpbGwtb3BhY2l0eToxO2ZpbGwtcnVsZTpub256ZXJvO3N0cm9rZTpub25lIi8+PC9nPjwvZz48L3N2Zz4=">';
```
jfirebaugh: SVG definitely won't work in IE 8, and I'm not sure about support for SVG data URLs in 9/10. Deserves testing and consideration about whether we care to support old IEs here.

camilleanne: @jfirebaugh did some browserstack testing -- IE 8 is the only one that broke. Thinking about how to get around this.
screen shot 2015-01-13 at 5 02 28 pm

camilleanne: "fixed" by moving image to css. Now it's not broken on IE 8, but just doesn't show.

src/map.js
```diff
@@ -77,6 +78,9 @@ var LMap = L.Map.extend({
             this.addControl(this.shareControl);
         }
 
+        this._mapboxLogoControl = mapboxLogoControl({urlOrMapID: _});
```
jfirebaugh: Instead of `{urlOrMapID: _}`, which is no longer used, pass `this.options.mapboxLogoControl` here, which will allow:

```
L.mapbox.map(map, mapid, {mapboxLogoControl: {position: 'topright'}});
```

src/mapbox_logo.js
```diff
@@ -0,0 +1,33 @@
...
+        position: 'bottomleft',
+    },
+
+    initialize: function(options) {
+        L.Util.setOptions(this, options);
+    },
+
+    onAdd: function(map) {
+        this._container = L.DomUtil.create('div', 'mapbox-logo');
+        L.DomEvent.disableClickPropagation(this._container);
```
jfirebaugh: What is click propagation and do we want to disable it?

camilleanne: Probably not needed since it's not a link with attached listeners -- it was copied over from another control.

PR #942, opened 2015-02-06
[don't merge] Responsive attribution

tristen:
Looks like this _(the gif is a little glitchy)_:

attribution-responsive

cc @jfirebaugh @lxbarth


jfirebaugh:
This needs to be responsive to the actual size of the map element rather than the window. Like if you have a small map in one corner of the page it should use the (i).

tristen:
@jfirebaugh https://github.com/mapbox/mapbox.js/commit/a1825c49765de666366840715bbf52e5415832d6 look right to you? 

Not sure the best place to document `compact:true` in https://github.com/mapbox/mapbox.js/blob/mb-pages/API.md

jfirebaugh:
> Not sure the best place to document compact:true

Add to the `L.mapbox.map` `options` list:
- `attributionControl`: value can be `{compact: true}` to force a compact attribution icon that shows the full attribution on click, or `{compact: false}` to force the full attribution control. The default is a responsive attribution that collapses when the map is less than 640 pixels wide.

jfirebaugh:
Note that we can make the responsive attribution changes in the next 2.x release, but removing infoControl support will need to wait for 3.0 (which will probably coincide with Leaflet 1.0).

lxbarth:
Can we deprecate the infoControl() API right away?

jfirebaugh:
We could mark it as deprecated in the documentation and `console.log` a warning if it's used. Is that what you had in mind @lxbarth?

lxbarth:
I think that makes sense right? This way we manage expectations that it'll go away in 3.

tristen:
@jfirebaugh @lxbarth 

So just in `map.js` something like this?

``` js
if (this.options.infoControl) {
    console.warn('infoControl has been deprecated in v3.0.0');
}
```

tristen:
-> https://github.com/mapbox/mapbox.js/pull/948

PR #948, opened 2015-02-12
Responsive attribution

tristen:
Re-roll of https://github.com/mapbox/mapbox.js/pull/942

cc @jfirebaugh


jfirebaugh:
Can you add tests for the `compact` option please?

tristen:
@jfirebaugh I'm trying to trigger a container change after asserting `.leaflet-compact-attribution` is present but the second test fails for me:

```js
describe('L.control.attribution', function() {
    'use strict';
    it('toggles leaflet-compact-attribution class', function() {
        var el = document.createElement('div');
        var map = L.mapbox.map(el, 'examples.map-zr0njcqy');
        var attributionControl = el.querySelector('.leaflet-control-attribution');
        expect(attributionControl.classList.contains('leaflet-compact-attribution')).to.eql(true);

        el.style.width = '641px';
        map.fire('resize');

        // Failing test.
        expect(attributionControl.classList.contains('leaflet-compact-attribution')).to.eql(false);
    });
});
```

I've tried `map.invalidateSize()` and setTimeout before the last test is executed but I get the same error . Running something like this in the browser works:

```js
var map = L.mapbox.map('map', 'examples.map-zr0njcqy');
map['_container'].style.width = '640px';
map.fire('resize');
console.log(document.querySelector('.leaflet-compact-attribution')); // returns false
```

test/spec/attribution_control.js
```diff
@@ -0,0 +1,29 @@
+describe('L.control.attribution', function() {
+    'use strict';
+    it('toggles leaflet-compact-attribution class', function() {
+        var el = document.querySelector('body').appendChild(document.createElement('div'));
```
jfirebaugh: The added element should be removed after the test completes, so that the DOM is clean for future tests. [example](https://github.com/mapbox/mapbox.js/blob/db25bba68d2fb55b549d2f23988cebc31a95bc5e/test/spec/share_control.js#L5-15)

test/spec/attribution_control.js
```diff
@@ -0,0 +1,29 @@
...
+        var attributionControl = el.querySelector('.leaflet-control-attribution');
+        expect(attributionControl.classList.contains('leaflet-compact-attribution')).to.eql(true);
+
+        el.style.width = '641px';
+        map.fire('resize');
+
+        expect(attributionControl.classList.contains('leaflet-compact-attribution')).to.eql(false);
+    });
+
+    it('persists leaflet-compact-attribution', function() {
```
jfirebaugh: `it('always adds the leaflet-compact-attribution class when {compact: true} is specified', ...)` and then add a test for `it('never adds the leaflet-compact-attribution class when {compact: false} is specified',...)`.

jfirebaugh:
:+1:

PR #949, opened 2015-02-15
Exporting images directory in bower.json

theorm:
Gulp assets pipeline is not able to see images directory because it is not in bower.json main section. Added it there.


tmcw:
:+1: Thanks! Will go out with the next release.

PR #959, opened 2015-02-27
Allow style to be configured for FeatureLayer

bnoon:
This allows setting the style of pointToLayer or L.Path layers.
The style option can be an object or a function that returns an object.
Similar to how L.geoJson works.

Tests are included.


jfirebaugh:
Thanks!

PR #960, opened 2015-02-28
Add FeatureLayer.setStyle and .getStyle methods

bnoon:
Need to save the style option so it can be used when resetting the layer.


bnoon:
Note that .setStyle() is documented in L.FeatureGroup

PR #969, opened 2015-03-12
jpg@2x

yhahn:
Adds support for modifying `tiles` URLs with the `.jpg` format to get @2xified automatically.

### DIFF
--- a/src/tile_layer.js
+++ b/src/tile_layer.js
@@ -73,7 +73,7 @@ var TileLayer = L.TileLayer.extend({
         if (!templated) {
             return templated;
         } else {
-            return templated.replace('.png',
+            return templated.replace(/\.(png|jpg)/,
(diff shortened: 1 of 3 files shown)


src/tile_layer.js
```diff
@@ -73,7 +72,7 @@ var TileLayer = L.TileLayer.extend({
         if (!templated) {
             return templated;
         } else {
-            return templated.replace('.png',
+            return templated.replace(/\.(png|jpg)/,
```
jfirebaugh: Use the [stricter regex from GL](https://github.com/mapbox/mapbox-gl-js/blob/fe072b7f178345abe94fecd7ab3e4b85bc77018d/js/util/mapbox.js#L64). This regex will fail with a hypothetical map id of `pwn.jpg` for instance.

yhahn: :+1:

Issue #955, opened 2015-02-18
Tooltip styling is broken using mapbox.standalone with leaflet


pixelwhip:
I'm using Leaflet 0.8-dev and Mapbox 2.1.5 

There seems to be conflicting style rules for the .leaflet-popup-tip in leaflet.css and mapbox.standalone.css that cause the tip to break. See the attached screenshot. 

![image](https://cloud.githubusercontent.com/assets/967665/6256029/090393e6-b772-11e4-9ce6-56b554783044.png)

jfirebaugh:
Yeah, it looks like the leaflet.js + leaflet.css + mapbox.standalone.js + mapbox.standalone.css combo will not work. mapbox.standalone.css is the same as mapbox.css, which is expected to be used in place of leaflet.css, not in addition to it.

@tristen - have I got this correct? If so, we should remove mapbox.standalone.css and write clear documentation on how to use mapbox.standalone.

pixelwhip:
@jfirebaugh Good to know. I did remove leaflet.css and everything seemed to work in our particular case, but wasn't sure what other effects that might have.

tristen:
@pixelwhip @jfirebaugh that looks to be the issue yep. Removing `mapbox.standalone.css` might be a good plan.

jfirebaugh:
Removing it is technically a breaking API change (some people are likely using leaflet.js + mapbox.standalone.js + mapbox.standalone.css), so let's wait until 3.0 for that. For now let's add a page to the documentation about using standalone mode.

Points to cover:
- This is for combining mapbox.js with a separate copy of leaflet, for example to use a different version than what's bundled, or a patched version
- Correct usage is leaflet.js + mapbox.standalone.js + mapbox.css (provide code snippet)
- Call out that Leaflet master / 0.8 / 1.0-beta is explicitly _not_ supported

davidr-halomicrosystems:
Yeah just experienced this using the following 👍 

v2.4.0/mapbox.js'
v2.4.0/mapbox.css'

and leaflet.css. Was worried I needed it as im using plugins/leaflet-fullscreen/v1.0.1/leaflet.fullscreen as well. Anyway removed leaflet.css and all is well now

PR #982, opened 2015-04-10
Add standalone mapbox.js info and change map id 

lyzidiamond:
This PR:
- Adds a box to the /mapbox.js page with info on mapbox.standalone.js, which is referred to in [our FAQ](https://www.mapbox.com/help/mapboxjs-a-leaflet-plugin/). Will change the language for that in the website repo.
- Removes the `examples` mapid from the /mapbox.js page and replaces it with `mapbox.streets`.

/cc @tmcw


jfirebaugh:
I think this is better done as a subsection in [the "Guides" section of API.md](https://github.com/mapbox/mapbox.js/blob/mb-pages/API.md#guides). From there, it will automatically be pulled into the online docs at a URL like https://www.mapbox.com/mapbox.js/api/v2.1.7/standalone/, which you can link to from https://www.mapbox.com/help/mapboxjs-a-leaflet-plugin/ rather than saying "go to page A, click B, then look for C".

lyzidiamond:
:+1: totally agree @jfirebaugh, will make that change

lyzidiamond:
Thanks for the advice @jfirebaugh! This is now updated.

jfirebaugh:
Should have noted this earlier, but I just remembered: we've deprecated mapbox.standalone.css, because it implies that you also need to supply leaflet.css, and that doesn't actually work. The correct usage is is leaflet.js + mapbox.standalone.js + mapbox.css. See https://github.com/mapbox/mapbox.js/issues/955 for more.

lyzidiamond:
Updated the docs here.

PR #992, opened 2015-04-27
Never apply simplestyle rules to circle or circlemarker instances

### LINKED ISSUE #991: Problem with L.circleMarker in pointToLayer with L.mapbox.featureLayer
When using `L.circleMarker` with `pointToLayer` in an `L.mapbox.featureLayer`, styling options defined in `L.circleMarker` are not reflected on the map. This issue persists in `2.1.8`, `2.1.7`, and `2.1.6`.

Code:

```
var myLayer = L.mapbox.featureLayer(geojson, {
  pointToLayer: function(feature, latlon) {
    return L.circleMarker(latlon, {
      fillColor: '#ff0000'
    });
  }
}).addTo(map);
```

Map output:

screen shot 2015-04-23 at 1 16 11 pm

The two PRs that added `pointToLayer` to `L.mapbox.featureLayer` were #938 and #959. Perhaps something happened in there to cause this issue?

tmcw:
Fixes #991 

This makes the rule that

**L.circleMarker or L.circle layers produced by a pointToLayer method are immune to any style overrides**

This is a somewhat narrow fix to #991 but I think it's safe & good because
- simplestyle-spec only applies to geometry primitives & raster markers. circles and circle markers do not fit into either category

/cc @lyzidiamond

### DIFF
--- a/src/feature_layer.js
+++ b/src/feature_layer.js
@@ -93,14 +93,16 @@ var FeatureLayer = L.FeatureGroup.extend({
                 },
                 layer = L.GeoJSON.geometryToLayer(json, pointToLayer),
                 popupHtml = marker.createPopup(json, this.options.sanitizer),
-                style = this.options.style;
+                style = this.options.style,
+                defaultStyle = style === simplestyle.style;
 
             if (style && 'setStyle' in layer &&
+                // if the style method is the simplestyle default, then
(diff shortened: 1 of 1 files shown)


lyzidiamond:
This is great @tmcw, thank you so much.

jfirebaugh:
@tmcw Can you add a test for this?

Issue #1003, opened 2015-05-15
Support JSPM package loading.


lostinplace:
As it stands using JSPM to load mapbox fails badly.  This is unfortunate for me, since I would like to use mapbox instead of vanilla leaflet (which works under JSPM)

I might be able to figure out how to fi this, but if you guys have ideas (or just suggestions to speed me up)  that would be really helpful

-Chris

tmcw:
Can you be more specific about what "fails badly" means, exactly, with error messages or results?

JSPM promises

> Load any module format (ES6, AMD, CommonJS and globals) directly from any registry such as npm and GitHub with flat versioned dependency management.

So it's likely that any bug here is on JSPM's side or a usage issue.

lostinplace:
Hey there tmcw

so I've installed mapbox using:

```
jspm install github:mapbox/mapbox.js
```

I have a javascript file that imports it using the following

```
import L from 'mapbox';
```

this yields the following errors:

```
Navigated to http://localhost:3800/map.html
2015-05-15 15:30:14.771 es6-module-loader.src.js:2612 GET http://localhost:3800/jspm_packages/github/mapbox/mapbox.js@2.1.9/package.json.js 404 (Not Found)f @ es6-module-loader.src.js:2612$__Object$defineProperty.value @ es6-module-loader.src.js:2787c @ es6-module-loader.src.js:377b @ es6-module-loader.src.js:365$__Object$defineProperty.value @ es6-module-loader.src.js:2786$__global.upgradeSystemLoader.e.fetch @ system.src.js:739$__global.upgradeSystemLoader.e.fetch @ system.src.js:1806$__global.upgradeSystemLoader.e.fetch @ system.src.js:1921(anonymous function) @ es6-module-loader.src.js:1525O @ es6-module-loader.src.js:1183K @ es6-module-loader.src.js:11427.y.when @ es6-module-loader.src.js:9307.v.run @ es6-module-loader.src.js:8213.a._drain @ es6-module-loader.src.js:973.a.drain @ es6-module-loader.src.js:62b @ es6-module-loader.src.js:268
[...]

PR #1004, opened 2015-05-15
Support module exports for JSPM

lostinplace:
this is a simple, slightly hacky solution, that makes JSPM support immediately available

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,2 +1,4 @@
 require('./leaflet');
 require('./mapbox');
+var module = module || {};
+module.exports = window.L;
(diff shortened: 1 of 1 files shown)


lostinplace:
this reference #1003

tmcw:
This is very hacky. Why does JSPM want this to be run in a function?

lostinplace:
jshint warned about not using ""use strict"  so I added a strict function to manage the initializers.

This has nothing to do with jspm  It's worth noting that this will also fix the browserify problem

tmcw:
Superseded by #1008

PR #1023, opened 2015-06-15
Prox [WIP]

ingalls:
Add proximity support to mapbox.js

### DIFF
--- a/src/geocoder.js
+++ b/src/geocoder.js
@@ -11,6 +11,8 @@ module.exports = function(url, options) {
     if (!options) options = {};
     var geocoder = {};
 
+    console.log("OPTIONS:", JSON.stringify(options));
+
     util.strict(url, 'string');
 
     if (url.indexOf('/') === -1) {
@@ -20,6 +22,7 @@ module.exports = function(url, options) {
             url = urlhelper('/geocode/' + url + '/{query}.json', options.accessToken);
         }
     }
+    console.log(url);
 
     geocoder.getURL = function() {
         return url;
--- a/src/geocoder.js
+++ b/src/geocoder.js
@@ -11,8 +11,6 @@ module.exports = function(url, options) {
     if (!options) options = {};
     var geocoder = {};
 
-    console.log("OPTIONS:", JSON.stringify(options));
-
     util.strict(url, 'string');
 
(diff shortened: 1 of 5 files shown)


src/geocoder.js
```diff
@@ -35,12 +40,16 @@ module.exports = function(url, options) {
...
-        return L.Util.template(geocoder.getURL(), {query: query});
+
+        return L.Util.template(geocoder.getURL(), {
+            proximity: args.proximity ? encodeURIComponent(args.proximity[0]+','+args.proximity[1]) : false,
+            query: query 
+        });
     };
 
-    geocoder.query = function(_, callback) {
+    geocoder.query = function(_, args, callback) {
```
jfirebaugh: [`query` is a public API](https://www.mapbox.com/mapbox.js/api/v2.1.9/l-mapbox-geocoder/#section-geocoder-query) -- we can't make a backwards incompatible change to the method signature until 3.0.

We could make the first parameter accept an object as well as a plain string, and support:

`geocoder.query({query: 'austin', proximity: latlng}, callback)`

WDYT?

src/geocoder_control.js
```diff
@@ -7,6 +7,7 @@ var GeocoderControl = L.Control.extend({
     includes: L.Mixin.Events,
 
     options: {
+        proximity: false,
```
jfirebaugh: Let's default this to `true`.

src/geocoder.js
ingalls: :+1:

src/geocoder_control.js
ingalls: To me that would be a breaking change to the API as one is going to get drastically different results than with it set to false. Would it be better to save this for `3.0.0`?

jfirebaugh: Hopefully drastically better results though, right? But, yeah, I can see an argument either way, and it's easy to switch later on. @tmcw want to be a tiebreaker?

tmcw: I don't think this justifies a 3.0.0: it's not a breaking change, it's an improvement. This fits as a minor version bump.

ingalls: :+1:

PR #1024, opened 2015-06-15
change toGeoJSON to getGeoJSON

lyzidiamond:
Fix example so it works!


katydecorah:
looks good @lyzidiamond!

PR #1025, opened 2015-06-16
Prox2

ingalls:
Add proximity support to mapbox.js


ingalls:
@sbma44 @tmcw @jfirebaugh Do someone mind giving this a code review?

src/geocoder.js
```diff
@@ -35,12 +40,23 @@ module.exports = function(url, options) {
         }
 
         feedback.record({geocoding: query});
-        return L.Util.template(geocoder.getURL(), {query: query});
+
+        return L.Util.template(geocoder.getURL(), {
+            proximity: args.proximity ? encodeURIComponent(args.proximity[0]+','+args.proximity[1]) : '',
```
tmcw: `args.proximity.join(',')`

API.md
```diff
@@ -316,17 +316,18 @@ A low-level interface to geocoding, useful for more complex uses and reverse-geo
 | Options | Value | Description |
 | ---- | ---- | ---- |
 | id _or_ url | string | Value must be <ul><li>A [geocoder index ID](https://www.mapbox.com/developers/api/geocoding/), e.g. `mapbox.places`</li><li>A geocoder API URL, like `{{site.tileApi}}/v4/geocode/mapbox.places/{query}.json`</li></ul> |
-| options | Object | The second argument is optional. If provided, it may include: <ul><li>`accessToken`: Mapbox API access token. Overrides `L.mapbox.accessToken` for this geocoder.</li></ul> |
+| options | Object | The second argument is optional. If provided, it may include: <ul><li>`accessToken`: Mapbox API access token. Overrides `L.mapbox.accessToken` for this geocoder.</li><li>`proximity`: boolean which enables viewport based proximity search</li></ul> |
 
 _Returns_ a `L.mapbox.geocoder` object.
 
-### geocoder.query(queryString, callback)
+### geocoder.query(queryString|options, callback)
```
tmcw, line 323: i think I prefer `geocoder.query(queryString, [options], callback)`, which follows on a bit closer to node apis

ingalls, line 323: @tmcw This is what I had originally but @jfirebaugh pointed out this is part of the public API and would therefore be a breaking change. In `3.0` I'd be in favour of switching to this.

src/geocoder.js
ingalls: :+1:

API.md
```diff
@@ -316,17 +316,18 @@ A low-level interface to geocoding, useful for more complex uses and reverse-geo
...
 
 Queries the geocoder with a query string, and returns its result, if any.
 
 | Options | Value | Description |
 | ---- | ---- | ---- |
 | queryString (_required_) | string | a query, expressed as a string, like 'Arkansas' |
+| options | object | an object containing the query and options parameters like `{ query: 'Austin', proximity: [lng, lat] }`
```
tmcw: I think this `proximity` object should be an object like `L.latLng` rather than a `[lng, lat]` array: L.latLng is the Leaflet expectation and it's what we handle elsewhere in this api, like in setView. Leaflet's expectations are the law of the land around here, whether we like latlon or not

sbma44:
@tmcw I did the `./_docs/build.sh v2.1.9` dance but my local copy of Jekyll still isn't showing the edits to API.md. Any suggestions for what might be going wrong?

tmcw:
Do you see the changes in git? I'm seeing the change made locally & the change in Jekyll

src/geocoder.js
```diff
@@ -8,20 +8,25 @@ var util = require('./util'),
...
-
+    if (!options) options = {};
     var geocoder = {};
 
     util.strict(url, 'string');
 
     if (url.indexOf('/') === -1) {
-        url = urlhelper('/geocode/' + url + '/{query}.json', options && options.accessToken);
+        if (options.proximity) {
+            url = urlhelper('/geocode/' + url + '/{query}.json?proximity={proximity}', options.accessToken);
```
jfirebaugh, line 19: This should be `?proximity=true`. The `{proximity}` syntax is for interpolating a value but here it's constant.

jfirebaugh, line 19: Uh, duh, I'm wrong, ignore.

src/geocoder.js
```diff
@@ -35,12 +40,23 @@ module.exports = function(url, options) {
...
+            query: query 
+        });
     };
 
     geocoder.query = function(_, callback) {
         util.strict(callback, 'function');
-        request(geocoder.queryURL(_), function(err, json) {
+
+        var args = {};
+        if (typeof _ === 'object' && !Array.isArray(_)) {
```
jfirebaugh: [`Array.isArray` is not supported by IE8](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility), which Mapbox.js supports. I suggest reversing the condition and testing for a string instead.

sbma44:
@tmcw not sure what wasn't working before but it's behaving itself now

src/geocoder.js
```diff
@@ -8,20 +8,25 @@ var util = require('./util'),
...
+            url = urlhelper('/geocode/' + url + '/{query}.json', options.accessToken);
+        }
     }
 
     geocoder.getURL = function() {
         return url;
     };
 
-    geocoder.queryURL = function(_) {
+    geocoder.queryURL = function(_, args) {
```
jfirebaugh: `queryURL` hasn't been documented as a public API, but I wouldn't be surprised if it's in use somewhere, and so long as we have to use the single argument format for `query`, let's use it for `queryURL` too. Meaning, keep the `queryURL` signature the same, move the type checking of `_` into `queryURL`, and leave `query` doing a straight pass through -- `geocoder.queryURL(_)`.

tmcw:
Lots of trailing spaces in this work, plz enable https://github.com/SublimeText/TrailingSpaces in the future :)

Grabbing this branch, hold off on commits for a bit.

ingalls:
I've installed https://github.com/bronson/vim-trailing-whitespace so should be good next time ;)

tmcw:
@ingalls @sbma44 this is now deployed as 2.2.0

PR #1035, opened 2015-07-12
Use svg instead of png to get better image quality

PeterDaveHello:


jfirebaugh:
Thanks for the suggestion -- looks like this was addressed in ad964c88ec709c21fc2916bddf394d79edf1a85e.

Issue #661, opened 2014-01-29
Broken link from L.mapbox.map page to L.map


jfirebaugh:
It links to .../l-map, but actual page is .../l-map-class!

See #654.

/cc @lyzidiamond

jfirebaugh:
The script that generates the links between the Mapbox.js and the Leaflet documentation assumes a particular convention for IDs, but the Leaflet documentation deviates from the convention [here](https://github.com/Leaflet/Leaflet/blob/d3435383e94e00c87e4988a05d45fde5e1e4223a/reference.html#L129), using `id="map-class"` rather than `id="map"`.

tmcw:
(as you'll soon find out, we also have an overeager CSS rule that makes anything with `#map` big and white and rectangular like a map)

lyzidiamond:
Ah, thanks for explaining! Was doing some Leaflet explaining at CfA and it was causing some problems for some folks.

mourner:
Yeah, I think I made an exception in Leaflet docs for the same reason.

geografa:
Tender link: http://support.mapbox.com/discussions/general-questions/44184-broke-links-in-api-docs

PR #1050, opened 2015-08-12
fix broken map-class links in docs

sbma44:
this guy:

image

### DIFF
--- a/docs/_posts/api/0200-01-01-v2.2.1-all.html
+++ b/docs/_posts/api/0200-01-01-v2.2.1-all.html
@@ -151,7 +151,7 @@ interactivity into your map, which you can easily access with <code><a href="#l-
 </table>
 <p><em>Example</em>:</p>
 <pre><code>// the second argument is optional
-var layer = L.mapbox.gridLayer(&#39;examples.map-20v6611k&#39;);
+var layer = L.mapbox.gridLayer(&#39;mapbox.light&#39;);
 </code></pre><p><em>Returns</em> a <code><a href="#l-mapbox-gridlayer">L.mapbox.gridLayer</a></code> object.</p>
 <p><em>Class</em>: <code><a href="#l-mapbox-gridlayer">L.mapbox.GridLayer</a></code></p>
 <h3 id="gridlayer-on">gridLayer.on(event, handler, context)</h3>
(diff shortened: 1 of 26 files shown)


jfirebaugh:
Thanks @sbma44. See #661 for backstory here. Two things:
- Link should be to `l-map-class` (`L.Map`) rather than l-mapbox-map (`L.mapbox.Map`).
- 0200-01-01-l-tilelayer.html is a generated file and will be overwritten if we regenerate the docs. The fix needs to be in the generation code, [here](https://github.com/mapbox/mapbox.js/blob/14d4632fc8613d07a039bf82ffd162a05492b05c/_docs/generate.js#L95) probably.

sbma44:
Thanks for the eyes, @jfirebaugh -- I was worried about not understanding the build process here, and seemingly with good reason. I'll get this building locally & figure out the appropriate fix.

sbma44:
chatted with @lyzidiamond about this -- she's going to try to catch this class of errors in a more comprehensive manner & we'll proceed with a fix from there

lyzidiamond:
@jfirebaugh @sbma44 Okay, finally fixed these!

boo regex yay regex <-- how i feel right now

sbma44:
Thanks so much for grabbing this. @jfirebaugh lmk if I should figure out how to build this for a final :eyes: ; guessing you might have the context to hit merge more quickly than I would.

docs/_posts/api/0200-01-01-v2.2.1-all.html
```diff
@@ -151,7 +151,7 @@ <h2 id="l-mapbox-gridlayer">L.mapbox.gridLayer(id|url|tilejson, options)</h2>
 </table>
 <p><em>Example</em>:</p>
 <pre><code>// the second argument is optional
-var layer = L.mapbox.gridLayer(&#39;mapbox.streets&#39;);
+var layer = L.mapbox.gridLayer(&#39;examples.map-20v6611k&#39;);
```
jfirebaugh: You didn't intend to revert back to an examples account map here, right?

lyzidiamond: well this whole example is wrong. `examples.map-20v6611k` is `mapbox.light` but neither of them are a `gridLayer()`... I will fix.

lyzidiamond: Reading through, these docs need some additional love w/r/t v3 and `examples`. I'm going to fix these instances and merge, but then go through and do some additional cleanup in a different PR.

lyzidiamond:
Per chat with @sbma44, I am going to merge here. We can roll back if we see anything glaring.

sbma44:
thank you so much for tackling this, @lyzidiamond

PR #1057, opened 2015-09-08
Adds support for types to geocoder.query()

liamwooding:
This PR adds support for types to geocoder.query() - types can be a string or an array of strings, passed in the same manner as `query` and `proximity`.

### DIFF
--- a/src/geocoder.js
+++ b/src/geocoder.js
@@ -40,6 +40,19 @@ module.exports = function(url, options) {
 
         var url = L.Util.template(geocoder.getURL(), {query: query});
 
+        if (isObject && _.types) {
+            if (isArray(_.types)) {
+                var typesString = '';
+                _.types.forEach(function (type, i) {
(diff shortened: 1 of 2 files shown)


src/geocoder.js
```diff
@@ -40,6 +40,19 @@ module.exports = function(url, options) {
 
         var url = L.Util.template(geocoder.getURL(), {query: query});
 
+        if (isObject && _.types) {
+            if (isArray(_.types)) {
+                var typesString = '';
+                _.types.forEach(function (type, i) {
```
tmcw: why not use `.join(',')` here?

liamwooding: Good question! Will amend.

tmcw:
:+1: thanks!

tmcw:
Merged in ede56efbde5a846d6824a2b0100acf0dc7e6cabb

PR #1059, opened 2015-09-10
add comment to circleMarker example re units

danswick:
Add a comment to clarify that circleMarker is drawn in pixels, not map units.


tmcw:
Merged in e42ba754b8d498ef7f6e61845630b77a03084cc3

PR #1063, opened 2015-10-06
Update geocoding version, remove global version numbering, round coordinates

heyitsgarrett:
We're no longer standardizing `v4` throughout all of our APIs, starting with geocoding, which moves from:

`../v4/geocode/..`

to

`../geocoding/v5/..`

In the interim, I've removed `v4` from the global `config.js` and am leaving it to each API call to include separately.

(I also made the `urlhelper` function name consistent throughout)

cc @tmcw @yhahn


tmcw:
urlhelper isn't a very good name for what this is. formatURL?

tmcw:
Also, the name of the file and the name of what you call its export should match.

heyitsgarrett:
Y, `urlHelper` was legacy so I didn't rock the boat too much. I've updated everything to `formatUrl` and got tests in order.

@tmcw @jfirebaugh - a final review would be great here before Mergeville, :us:

jfirebaugh:
Convention for multi-word file names is underscore, so it should be `format_url.js` rather than `formatUrl.js`.

heyitsgarrett:
:+1: @jfirebaugh

sbma44:
Added coordinate rounding & associated tests. Good to merge when you're comfortable, @tmcw / @jfirebaugh.

PR #1070, opened 2015-10-21
Set geocoding result title attribute. Fixes #1033

### LINKED ISSUE #1033: Truncated geocoder results should be available as title= attributes
Hovering here should give you the full value

tmcw:
cc @jfirebaugh for the review, this helps with the usability of
truncated geocoding results


jfirebaugh:
:+1:

Issue #1074, opened 2015-10-31
v2.2.3 would introduce breaking change


karenzshea:
Changes from https://github.com/mapbox/mapbox.js/pull/1063 will break uses of [mapbox.config.HTTP_URL](https://www.mapbox.com/mapbox.js/api/v2.2.2/l-mapbox-config-http_url/) and HTTPS_URL where a version number is set in the path. Most commonly seen in cahoots with api-maps-staging like in mapbox.js examples https://github.com/mapbox/mapbox.js/blob/1c6c03d5b5362b42a547d97a14417475f894b70c/docs/_layouts/example.html#L35

We could either add code to dedupe version paths or go for the 3 pointer

cc @yhahn @tmcw @sbma44

jfirebaugh:
Great catch @karenzshea! I like your approach to fixing this in #1076.

tmcw:
Fixed in #1076

PR #1076, opened 2015-11-02
Dedupe /v4 out of config url if manually set

karenzshea:
Introduced by https://github.com/mapbox/mapbox.js/pull/1063

Related ticket: https://github.com/mapbox/mapbox.js/issues/1074

This blocks rolling a new mapbox.js with geocoder updates and upgrading mapbox.js in atlas

@tmcw wanna review and help me deploy. maybe I should add a test for this?

### DIFF
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -12,8 +12,7 @@ module.exports = function(path, accessToken) {
     }
 
     var url = (document.location.protocol === 'https:' || config.FORCE_HTTPS) ? config.HTTPS_URL : config.HTTP_URL;
-    var vDupe = url.match('\/v[0-9]');
-    url = (vDupe !== null) ? url.split(vDupe[0])[0] : url;
+    url = url.replace(/\/v4$/, '');
     url += path;
     url += url.indexOf('?') !== -1 ? '&access_token=' : '?access_token=';
 
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -12,7 +12,8 @@ module.exports = function(path, accessToken) {
     }
 
     var url = (document.location.protocol === 'https:' || config.FORCE_HTTPS) ? config.HTTPS_URL : config.HTTP_URL;
-    url = (url.indexOf('/v4')) ? url.split('/v4')[0] : url;
+    var vDupe = url.match('\/v[0-9]');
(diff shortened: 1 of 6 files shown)


src/format_url.js
```diff
@@ -12,6 +12,7 @@ module.exports = function(path, accessToken) {
     }
 
     var url = (document.location.protocol === 'https:' || config.FORCE_HTTPS) ? config.HTTPS_URL : config.HTTP_URL;
+    url = (url.indexOf('/v4')) ? url.split('/v4')[0] : url;
```
yhahn: - Could this regex be a bit more permissive, e.g. in case anyone's pointing at v1-3?
- Are there any unit tests for this lib that could prove this code works?

jfirebaugh: - We have unit tests [here](https://github.com/mapbox/mapbox.js/blob/mb-pages/test/spec/format_url.js).
- The check here should be limited to `/v4` at the start of the path. We haven't supported `/v3` since 2.0.0; I don't think we need to start now.
- In case somebody is reading them, we should also revert `config.HTTP_URL` / `config.HTTPS_URL` to the previous values.

karenzshea:
@yhahn @jfirebaugh I added a test to `src/format_url.js` jfyi this test was never getting run by travis or included in the mocha-phantom runs until now.

@jfirebaugh clarify what you mean by?

> The check here should be limited to /v4 at the start of the path.  

The way I've written it [now](https://github.com/mapbox/mapbox.js/pull/1076/files#diff-6fed1961dd4993db51de9871fc1689e7R16) if there is any version added into the HTTP/S_URL path it will get dropped.

src/format_url.js
```diff
@@ -12,6 +12,8 @@ module.exports = function(path, accessToken) {
     }
 
     var url = (document.location.protocol === 'https:' || config.FORCE_HTTPS) ? config.HTTPS_URL : config.HTTP_URL;
+    var vDupe = url.match('\/v[0-9]');
```
jfirebaugh: Specifically, I'd replace these two lines with:

```
url = url.replace(/\/v4$/, '');
```

Only match `/v4`, only at the end of the string.

jfirebaugh:
> jfyi this test was never getting run by travis or included in the mocha-phantom runs until now

Good catch (again). :beers:

PR #1091, opened 2015-12-08
Fix "disable zooming and panning" example

lyzidiamond:
User pointed out that this line was missing from the example. cc @jfirebaugh @tmcw


jfirebaugh:
:+1:

Issue #878, opened 2014-09-17
Legend list items no longer unstyled


wboykinm:
Looks like your parent CSS for `<li>` elements must have changed; [the choropleth example](https://www.mapbox.com/mapbox.js/example/v1.0.0/choropleth/) has a bulleted legend:
![2014-09-17_1330](https://cloud.githubusercontent.com/assets/735463/4308208/a15f6080-3e90-11e4-87cd-58084b9e2384.png)

jfirebaugh:
Probably was implicitly depending on base CSS prior to the [iframening](https://github.com/mapbox/mapbox.js/issues/846).

PR #1101, opened 2016-01-12
fix CSS in Choropleth legend example

katydecorah:
Per #878 and a HS ticket, I updated the CSS to turn off the list-style:

image


jfirebaugh:
:+1:

PR #1104, opened 2016-01-14
Do not update geocoder control results on special keys

kirach:
A small change that prevents geocoder control from autocompleting (and update and re-render results) when special key (like escape, tab, enter, arrow keys) was pressed.

### DIFF
--- a/src/geocoder_control.js
+++ b/src/geocoder_control.js
@@ -187,7 +187,7 @@ var GeocoderControl = L.Control.extend({
     },
 
     _autocomplete: function(e) {
-        if (this.specialKeyCodes.indexOf(e.keyCode) != -1) return;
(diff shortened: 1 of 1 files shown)


kirach:
Guys, any feedback please? Do I need to change/add/remove something?

jfirebaugh:
I wonder if we can use the `input` event instead. That would avoid needing to use `keyCode` which is non-standard, deprecated, and system and implementation dependent.

/cc @tmcw

tmcw:
IE compatibility for input is pretty [sketchy according to the MDN docs](https://developer.mozilla.org/en-US/docs/Web/Events/input) - no support for IE<9, and IE9 doesn't fire input when characters are deleted, which is a pretty common occurrence here.

PR #1108, opened 2016-01-20
Add L.mapbox.styleLayer

bsudekum:
This adds the ability to add a style id in place of a mapid and get raster tiles of that style. Some notes:
- tiles must always be configured with [these options](https://github.com/mapbox/mapbox.js/compare/styles#diff-e1b538b03ec4f0f4fdced9f144bf186bR37)
- I added the ability to add the whole style object in place of the id [here](https://github.com/mapbox/mapbox.js/compare/styles#diff-e1b538b03ec4f0f4fdced9f144bf186bR64). I'm not sure if this makes sense because it will not be configurable at run time.
- The name `styleLayer` is still up for debate

Tests are failing because options are not properly set, any insight on this would be appreciated.

/cc @jfirebaugh @tmcw

### DIFF
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -41,3 +41,14 @@ module.exports.tileJSON = function(urlOrMapID, accessToken) {
 
     return url;
 };
+
+
+module.exports.style = function(styleURL, accessToken) {
+    if (styleURL.indexOf('mapbox://styles/') === -1) throw new Error('Incorrectly formatted Mapbox style at ' + styleURL);
+
+    var ownerIDStyle = styleURL.split('mapbox://styles/')[1];
+    var url = module.exports('/styles/v1/' + ownerIDStyle, accessToken)
+        .replace('http://', 'https://');
+
+    return url;
+};
--- a/src/style_layer.js
+++ b/src/style_layer.js
@@ -1,0 +1,81 @@
+'use strict';
+
+var util = require('./util');
+var format_url = require('./format_url');
+var request = require('./request');
+
+var StyleLayer = L.TileLayer.extend({
+
+    options: {
+        sanitizer: require('sanitize-caja')
+    },
+
+    initialize: function(_, options) {
+        L.TileLayer.prototype.initialize.call(this, undefined, options);
+
+        this.options.tiles = this._formatTileURL(_);
+        this.options.tileSize = 512;
+        this.options.zoomOffset = -1;
+        this.options.tms = false;
+
+        this._getAttribution(_);
+    },
+
+    _getAttribution: function(_) {
+        var styleURL = format_url.style(_, this.options && this.options.accessToken);
+        request(styleURL, L.bind(function(err, style) {
+            if (err) {
+                util.log('could not load Mapbox style at ' + styleURL);
+                this.fire('error', {error: err});
+            }
+            var sources = [];
+            for (var id in style.sources) {
+                var source = style.sources[id].url.split('mapbox://')[1];
+                sources.push(source);
+            }
+            request(format_url.tileJSON(sources.join(), this.options.accessToken), L.bind(function(err, json) {
+                if (err) {
+                    util.log('could not load TileJSON at ' + _);
+                    this.fire('error', {error: err});
+                } else if (json) {
+                    util.strict(json, 'object');
+
+                    this.options.attribution = this.options.sanitizer(json.attribution);
+
+                    this._tilejson = json;
+                    this.fire('ready');
+                }
+            }, this));
+        }, this));
+    },
+
+    // disable the setUrl function, which is not available on mapbox tilelayers
+    setUrl: null,
+
+    _formatTileURL: function(style) {
+        var retina = L.Browser.retina ? '@2x' : '';
+        if (typeof style === 'string') {
+            if (style.indexOf('mapbox://styles/') === -1) {
+                util.log('Incorrectly formatted Mapbox style at ' + style);
+                this.fire('error');
+            }
+            var ownerIDStyle = style.split('mapbox://styles/')[1];
+            return format_url('/styles/v1/' + ownerIDStyle + '/tiles/{z}/{x}/{y}' + retina, this.options.accessToken);
+        } else if (typeof style === 'object') {
+            return format_url('/styles/v1/' + style.owner + '/' + style.id + '/tiles/{z}/{x}/{y}' + retina, this.options.accessToken);
+        }
+    },
+
+    // this is an exception to mapbox.js naming rules because it's called
+    // by `L.map`
+    getTileUrl: function(tilePoint) {
+        var templated = L.Util.template(this.options.tiles, tilePoint);
+        return templated;
+    }
+});
+
+module.exports.StyleLayer = StyleLayer;
+
+module.exports.styleLayer = function(_, options) {
+    return new StyleLayer(_, options);
+};
--- a/test/spec/format_url.js
+++ b/test/spec/format_url.js
@@ -48,4 +48,10 @@ describe("format_url", function() {
             expect(internals.url.tileJSON('user.map')).to.equal('https://a.tiles.mapbox.com/v4/user.map.json?access_token=key&secure');
         });
     });
+
+    describe('.style', function() {
+        it('returns a style url with access_token parameter', function() {
+            expect(internals.url.style('mapbox://styles/bobbysud/cifr15emd00007zlzxjew2rar')).to.equal('https://a.tiles.mapbox.com/styles/v1/bobbysud/cifr15emd00007zlzxjew2rar?access_token=key')
+        });
+    });
 });
--- a/test/spec/style_layer.js
+++ b/test/spec/style_layer.js
@@ -1,0 +1,108 @@
+describe('L.mapbox.styleLayer', function() {
+    var server;
+
+    beforeEach(function() {
+        server = sinon.fakeServer.create();
+    });
+
+    afterEach(function() {
+        server.restore();
+    });
+
+    describe('Constructor', function() {
+
+        it('sets attribution', function(done) {
+            var layer = L.mapbox.styleLayer('mapbox://styles/mapbox/streets-v8');
+            layer.on('ready', function(e) {
+                expect(layer.options.attribution).to.equal('<a href="https://www.mapbox.com/about/maps/" target="_blank">&copy; Mapbox</a> <a href="http://www.openstreetmap.org/about/" target="_blank">&copy; OpenStreetMap</a> <a class="mapbox-improve-map" href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a>');
+                done();
+            });
+
+            server.respondWith('GET', 'https://a.tiles.mapbox.com/styles/v1/mapbox/streets-v8?access_token=key',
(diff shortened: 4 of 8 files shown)


src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+        return this._tilejson;
+    },
+
+    // disable the setUrl function, which is not available on mapbox tilelayers
+    setUrl: null,
+
+    _formatTileURL: function(style) {
+        var retina = L.Browser.retina ? '@2x' : '';
+        if (typeof style === 'string') {
+            if (style.indexOf('mapbox://styles/') === -1) {
```
tmcw, line 58: there's non-mapbox:// support in the constructor, but not here: why?

src/load_tilejson.js
```diff
@@ -7,16 +7,40 @@ var request = require('./request'),
...
+                    if (err) {
+                        util.log('could not load Mapbox style at ' + styleURL);
+                        this.fire('error', {error: err});
+                    }
+                    var sources = [];
+                    for (var id in style.sources) {
+                        var source = style.sources[id].url.split('mapbox://')[1];
+                        sources.push(source);
+                    }
+                    request(format_url.tileJSON(sources.join(), this.options.accessToken), L.bind(function(err, json) {
```
tmcw: i'm confused why we request the tilejson for every source; is that needed?

bsudekum: @tmcw  This is not making a request for every source, just one source. the tilejson endpoint supports concatenated sources, ex:

```
https://api.mapbox.com/v4/mapbox.mapbox-streets-v6,mapbox.satellite.json?access_token=pk.eyJ1IjoiYm9iYnlzdWQiLCJhIjoiTi16MElIUSJ9.Clrqck--7WmHeqqvtFdYig&secure
```

src/style_layer.js
bsudekum, line 58: > there's non-mapbox:// support in the constructor

Where is this happening? It should not, `styleLayer` should only with Mapbox.

src/load_tilejson.js
tmcw: Okay, but the same question: why is this requesting source JSON at all?

bsudekum: Attribution

lyzidiamond:
I can write an example for this once it's merged, unless you already have one @bsudekum

bsudekum:
@lyzidiamond go for it :+1:

src/format_url.js
```diff
@@ -41,3 +41,14 @@ module.exports.tileJSON = function(urlOrMapID, accessToken) {
...
     return url;
 };
+
+
+module.exports.style = function(styleURL, accessToken) {
+    if (styleURL.indexOf('mapbox://styles/') === -1) throw new Error('Incorrectly formatted Mapbox style at ' + styleURL);
+
+    var ownerIDStyle = styleURL.split('mapbox://styles/')[1];
+    var url = module.exports('/styles/v1/' + ownerIDStyle, accessToken)
+        .replace('http://', 'https://');
```
bsudekum, line 51: `HTTPS` is required for this endpoint.

src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+            minZoom: json.minzoom || 0,
+            maxZoom: json.maxzoom || 22,
+            bounds: json.bounds && util.lbounds(json.bounds),
+            tms: false,
+            tileSize: 512,
+            zoomOffset: -1
+        });
+
+        this._tilejson = json;
+        this.redraw();
```
bsudekum: `this.redraw();` can cause some flashing if fetching the tilesjson takes longer than usual.

bsudekum:
@tmcw @jfirebaugh I fixed up the tests. Only remaining task is https://github.com/mapbox/mapbox.js/pull/1108#discussion_r50266593

Any other feedback/ :eyes:  would be appreciated.

src/load_tilejson.js
```diff
@@ -7,16 +7,40 @@ var request = require('./request'),
...
-            request(_, L.bind(function(err, json) {
-                if (err) {
-                    util.log('could not load TileJSON at ' + _);
-                    this.fire('error', {error: err});
-                } else if (json) {
-                    this._setTileJSON(json);
-                    this.fire('ready');
-                }
-            }, this));
+            if (_.indexOf('mapbox://') > -1) {
```
jfirebaugh: This code should live in style_layer.js. This function is called `loadTileJSON` -- it shouldn't gain responsibilities beyond that.

src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+        return this;
+    },
+
+    getTileJSON: function() {
```
jfirebaugh: Remove this; we might remove the internal need for `styleLayer` to load TileJSON in the future, so we don't want to commit to making it part of the public API.

src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+var format_url = require('./format_url');
+
+var StyleLayer = L.TileLayer.extend({
+    includes: [require('./load_tilejson')],
+
+    options: {
+        sanitizer: require('sanitize-caja')
+    },
+
+    scalePrefix: '@2x',
```
jfirebaugh: Unused?

src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+
+    initialize: function(_, options) {
+        L.TileLayer.prototype.initialize.call(this, undefined, options);
+
+        this.options.tiles = this._formatTileURL(_);
+
+        this._tilejson = {};
+
+        if (options && options.format) {
```
jfirebaugh: Unused?

src/style_layer.js
```diff
@@ -0,0 +1,88 @@
...
+
+        this._loadTileJSON(_);
+    },
+
+    _setTileJSON: function(json) {
+        util.strict(json, 'object');
+
+        L.extend(this.options, {
+            attribution: this.options.sanitizer(json.attribution),
+            minZoom: json.minzoom || 0,
```
jfirebaugh: Let's not use any attributes from the TileJSON other than `attribution`.

bsudekum:
@jfirebaugh :+1: thanks, updated with feedback

src/style_layer.js
```diff
@@ -0,0 +1,100 @@
...
+    // by `L.map`
+    getTileUrl: function(tilePoint) {
+        var templated = L.Util.template(this.options.tiles, tilePoint);
+        return templated;
+    },
+
+    // TileJSON.TileLayers are added to the map immediately, so that they get
+    // the desired z-index, but do not update until the TileJSON has been loaded.
+    _update: function() {
+        if (this.options.tiles) {
```
jfirebaugh: Since the constructor sets `this.options.tiles`, this condition will always be true.

Instead, let's remove the asynchronicity around tile properties, and figure out how to refresh just the attribution when we get it:
- Move setting `tileSize: 512` and `zoomOffset: -1` to the constructor.
- Inline the rest of `_setTileJSON` into `_getAttribution`.
- Figure out what needs to happen inside `_getAttribution` to refresh the attribution when we get it.
- Remove `_update`.

src/style_layer.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
+
+var util = require('./util');
+var format_url = require('./format_url');
+var request = require('./request');
+
+var StyleLayer = L.TileLayer.extend({
+    includes: [require('./load_tilejson')],
```
jfirebaugh: No longer necessary, I believe.

src/style_layer.js
```diff
@@ -0,0 +1,100 @@
...
+
+        this._getAttribution(_);
+    },
+
+    _setTileJSON: function(json) {
+        util.strict(json, 'object');
+
+        L.extend(this.options, {
+            attribution: this.options.sanitizer(json.attribution),
+            tms: false,
```
jfirebaugh: `tms: false` is the default.

src/style_layer.js
jfirebaugh: This will be fixed by https://github.com/mapbox/mapbox.js/pull/1108/files#r50455207. `this.redraw()` removes and readds all tiles, but we only need to refresh attribution.

src/style_layer.js
bsudekum: The intention of adding this is to make sure it cannot be `true`, ever.

bsudekum:
Updated, and removed the redraw function. Flickering no longer occurs.

src/style_layer.js
jfirebaugh: There are many [`L.TileLayer` options](http://leafletjs.com/reference.html#tilelayer) that will cause this class to malfunction if they are misused. Why call out `tms` specifically?

src/style_layer.js
```diff
@@ -0,0 +1,83 @@
...
+
+    initialize: function(_, options) {
+        L.TileLayer.prototype.initialize.call(this, undefined, options);
+
+        this.options.tiles = this._formatTileURL(_);
+        this.options.tileSize = 512;
+        this.options.zoomOffset = -1;
+        this.options.tms = false;
+
+        this._tilejson = {};
```
jfirebaugh: This is now never read from and can be removed.

test/spec/style_layer.js
```diff
@@ -0,0 +1,90 @@
...
+        beforeEach(function() {
+            retina = L.Browser.retina;
+            L.Browser.retina = false;
+        });
+
+        afterEach(function() {
+            L.Browser.retina = retina;
+        });
+
+        it('distributes over the URLs in the tiles property', function() {
```
jfirebaugh: "distributes over the URLs in the tiles property" refers to the host-sharding behavior over TileJSON `tiles` properties. This implementation always uses `a.tiles.mapbox.com`, so the test isn't testing what it claims to be testing.

We should probably change the implementation to shard across `a.tiles.mapbox.com` and `b.tiles.mapbox.com`, for consistency with how `L.mapbox.tileLayer` will behave given that our current TileJSON responses include those two hosts in `tiles`.

test/spec/format_url.js
```diff
@@ -48,4 +48,10 @@ describe("format_url", function() {
             expect(internals.url.tileJSON('user.map')).to.equal('https://a.tiles.mapbox.com/v4/user.map.json?access_token=key&secure');
         });
     });
+
+    describe('.style', function() {
+        it('returns the input when passed a URL', function() {
```
jfirebaugh: This isn't an accurate description of this test.

test/spec/style_layer.js
```diff
@@ -0,0 +1,90 @@
+describe('L.mapbox.styleLayer', function() {
+    var server;
+
+    beforeEach(function() {
+        server = sinon.fakeServer.create();
```
jfirebaugh, line 5: You should install fake responses with known values for URLs that the tests in this file request, so that the tests don't break if we change the attributions for `streets-v8`, `hybrid-streets-v8`, etc.

test/spec/style_layer.js
jfirebaugh: On second thought, we don't want to hard code `a.tiles.mapbox.com` and `b.tiles.mapbox.com`, so we only have `config.HTTPS_URL`. Let's leave the current behavior but change the test description to be more accurate.

bsudekum:
:+1: updated with feedback

jfirebaugh:
:+1: Squash down to one commit and merge!

PR #1113, opened 2016-01-27
Package.json: Upgrade mustache

jmcelroy5:
v2.2.1 fixes a content injection vulnerability

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "corslite": "0.0.6",
     "isarray": "0.0.1",
     "leaflet": "0.7.7",
-    "mustache": "0.7.3",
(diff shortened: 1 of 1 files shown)


alulsh:
Thanks @jmcelroy5 for this security fix. I brought this up to @tmcw shortly after you launched this PR and he does not believe mapbox.js is currently vulnerable due to our use of [sanitize-caja](https://github.com/mapbox/sanitize-caja) for all templated data and user-provided input.

Nevertheless, we should still upgrade mustache per your PR. I tested upgrading mustache v2.2.1 and it passed all of our tests. This slated to be included in mapbox.js version 2.2.5, which will be released soon.

/cc @mapbox/security

PR #1118, opened 2016-02-21
Re-compressed PNGs for smaller size

nyurik:
Pushing upstream size optimization done by Ori
Patch: https://gerrit.wikimedia.org/r/#/c/272115/
Info:  https://phabricator.wikimedia.org/T127608

Zopfli is the most efficient DEFLATE compression algorithm, trading run-time
performance for file sizes that are typically 3-8% smaller than those produced
by zlib with the maximum compression setting. Its output is Deflate-compatible,
so no specialized decoder is needed.
This change was created by running zopflipng against all the PNG files in this
repository. The exact invocation was:
  git ls-files --exclude-per-directory=.gitignore -- '*.png' |
    parallel zopflipng -m -y {} {} \;
Files which zopflipng was not able to compress more efficiently were left unmodified.


nyurik:
Resubmitted as #1127

PR #1122, opened 2016-03-01
mapbox.js dupe meta descriptions

zachbeattie:
per https://github.com/mapbox/business/issues/2924

### DIFF
--- a/_docs/generate.js
+++ b/_docs/generate.js
@@ -27,8 +27,8 @@ output.write(header);
 output.write('version: ' + argv.t + '\n');
 landOutput.write('version: ' + argv.t + '\n');
 
-output.write('description: Build anything with Mapbox.js, a library for fast & interactive maps.\n');
-landOutput.write('description: Build anything with Mapbox.js, a library for fast & interactive maps.\n');
+output.write('description: Build anything with Mapbox.js ' + argv.t + ', a library for fast & interactive maps.\n');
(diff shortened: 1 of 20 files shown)


zachbeattie:
@pdgoodman - eyes before I merge? This adds the mapbox.js version # to each meta description that had a duplicate.

jfirebaugh:
These are all generated files; we should update the [generation code](https://github.com/mapbox/mapbox.js/blob/b656f19fd413380b5f88464eccda20b587c85b3b/_docs/generate.js#L30-L31) to include the version number. Otherwise new versions won't get a version number, and these changes will get overwritten if we ever regenerate the files.

zachbeattie:
> we should update the generation code to include the version number. 

Good call @jfirebaugh, although I'm not sure how to make this update. Is this something you can help with?

jfirebaugh:
Sure! Just above and below the lines I linked to you can see examples of how to concatenate the version (in `argv.t`) into other strings. So including it in the description too is a matter of inserting `' + argv.t + '` in the right place in those two lines. You can test that it works by following [these instructions](https://github.com/mapbox/mapbox.js/blob/mb-pages/DOCUMENTING.md#api-documentation) (which in turn assume that you've previously run `npm install`).

zachbeattie:
Awesome, thanks @jfirebaugh. I updated and tested the generation code to include the version number.

PR #1120, opened 2016-02-27
Add Styles API example for vanilla leaflet

danswick:
An example demonstrating how to use Styles API tiles by setting tile size and zoom offset in Leaflet. 

cc: @bsudekum @lyzidiamond

PR #1124, opened 2016-03-02
Update leaflet attribution to latest

danswick:
Per https://github.com/mapbox/mapbox.js/pull/1120#issuecomment-190973961, updates Leaflet attribution to be in line with https://www.mapbox.com/mapbox.js/example/v1.0.0/plain-leaflet/. 

cc: @jfirebaugh


jfirebaugh:
:+1: Thank you!

Issue #1126, opened 2016-03-13
Bug in geometryToLayer() call


nyurik:
FeatureLayer calls `geometryToLayer(json, options)` with a function as the second parameter, instead of passing in an object with that function as one of the values. Compare:

[mapbox code](https://github.com/mapbox/mapbox.js/blob/95a08fc8f8a7590a74698a1d53cd9222897159f3/src/feature_layer.js#L94)

[leaflet code](https://github.com/Leaflet/Leaflet/blob/7de33f0f7b4191e1ff92dbf838856eab51196c78/src/layer/GeoJSON.js#L76)

nyurik:
After a few trial and errors, I finally realized that mapbox.js was using leaflet 0.7.7, whereas the [leaflet-headless](https://github.com/jieter/leaflet-headless) i was trying to use was already on leaflet 1.0.0beta.  Any plans/etca to upgrade?

tmcw:
Yes, in https://github.com/mapbox/mapbox.js/issues/1038#issuecomment-137818756 - as soon as 1.0 is released, we'll track it.

nyurik:
thx!  @tmcw, could you check pull #1127 - its a very simple PNG recompression to optimize its size.

PR #1128, opened 2016-03-13
Pass pointToLayer() to geometryToLayer() in FeatureLayer

nyurik:
See issue #1126


nyurik:
silly me, mapbox.js is using leaflet 0.7.7, whereas this is the leaflet 1.0.0beta syntax. Hope mapbox is updated soonish.

PR #1130, opened 2016-03-25
Support autocomplete and country geocoder options.

### LINKED ISSUE #1129: Support 'country' option on Geocoder and geocoder API methods
We now support a few great extra methods for our geocoder: we should support them here too!

tmcw:
Fixes #1129 cc @jfirebaugh for the review


jfirebaugh:
:+1:

PR #1132, opened 2016-03-30
Throw an error if someone tries to use a Mapbox Studio style.

### LINKED ISSUE #1071: Add a friendly error for user to try to use a GL style id
Mapbox GL Style IDs take the form `mapbox://styles/[user]/[id]`. Since they can only be used within mapbox-gl-js and not within mapbox.js, let's display a friendly error message if a user tries to use one.

tmcw:
Fixes #1071 

cc @scothis for the review

### DIFF
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -29,6 +29,12 @@ module.exports = function(path, accessToken) {
 };
 
 module.exports.tileJSON = function(urlOrMapID, accessToken) {
+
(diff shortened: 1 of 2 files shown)


scothis:
Would it be better to refer them to `L.mapbox.styleLayer`?

tmcw:
Yep, good point. changed the message.

scothis:
LGTM

PR #1138, opened 2016-04-19
Use SVGs with PNG fallbacks for toolbar icons

### LINKED ISSUE #1137: Use SVG for icons
Currently we build a PNG and a 2x PNG. In most browsers we can just use the SVG directly.

edg2s:
Simplify the SVG source as it is now being served to the
client. The Inkscape metadata contained nothing of use.

With the simplified SVG we can now create the color variants
using a simple text replace in the SVG.

These SVGs can then be used to create the colored PNGs directly
which removes the dependency on ImageMagick.

Fixes #1137.


tmcw:
Sorry, [Mapbox.js supports IE8+](https://www.mapbox.com/mapbox.js/api/v2.4.0/) and [IE8 does not support SVG](http://caniuse.com/#feat=svg).

edg2s:
As per the discussion on the linked task this does not remove PNGs or IE8 support. Please re-open.

edg2s:
@tristen @tmcw Can someone please re-open this patch as it does not affect IE8 support as claimed.

edg2s:
Bump.

edg2s:
Are there any more objections to this? I'm wondering why it is taking so long to get merged?

tmcw:
Sorry about the delay, was hoping to get some input from @tristen and the folks who worked on the original icon set. Merging.

edg2s:
Thanks

Issue #1160, opened 2016-09-06
TypeError: Cannot read property 'sources' of null


rarkins:
Mapbox version: 2.4.0
Leaflet version: 0.7.7

I am using a client-side exception-catching service, and see a small percentage of cases with the following exception:

```
TypeError: Cannot read property 'sources' of null
  File "node_modules/mapbox.js/src/style_layer.js", line 32, in e.<anonymous>
                for (var id in style.sources) {
  File "node_modules/leaflet/dist/leaflet-src.js", line 54, in [anonymous]
                        return fn.apply(obj, args || arguments);
  File "node_modules/mapbox.js/src/request.js", line 28, in XMLHttpRequest.r
            callback(err, resp);
  File "node_modules/corslite/corslite.js", line 65, in XMLHttpRequest.r.onerror
            callback.call(this, evt || true, null);
```

The impacted code corresponds to https://github.com/mapbox/mapbox.js/blob/mb-pages/src/style_layer.js#L26-L35:

``` javascript
        request(styleURL, L.bind(function(err, style) {
            if (err) {
                util.log('could not load Mapbox style at ' + styleURL);
                this.fire('error', {error: err});
            }
            var sources = [];
            for (var id in style.sources) {
                var source = style.sources[id].url.split('mapbox://')[1];
                sources.push(source);
            }
```

I _do not_ see any 'could not load Mapbox style' log from the users prior to the null sources error, so either that `util.log` is suppressed or perhaps `err` is null?

jfirebaugh:
cc @bsudekum

bsudekum:
Looking at the code, it looks like we're not safe guarding against styles with no sources.

Will put together a test case for this.

tmcw:
The code also doesn't exit early if an error is found, so even if `err` is defined and `style` is null, it'll still try to iterate over them. @bsudekum do you have bandwidth to get a fix in?

PR #1170, opened 2016-10-12
Fix styleLayer with no style response

bsudekum:
Closes: https://github.com/mapbox/mapbox.js/issues/1160

In some cases, a style is requested but an error nor style is returned. In this case, we should bail early and not try to render the map. 

Couple of thoughts:
1. the `err` object returned will be null. This seems a little funny, I guess we just need error to fire. We're following this pattern in a [few places](https://github.com/mapbox/mapbox.js/blob/mb-pages/src/geocoder_control.js#L104-L105). Happy to construct an error object if necessary. 
2. It'd be great to find a map that consistently fails as I'm curious as to how we can not return a map and also an error. Only thing I can think of is a backend error like `504` that might slip by the `if err` check.

/cc @jfirebaugh @tmcw

### DIFF
--- a/src/style_layer.js
+++ b/src/style_layer.js
@@ -24,28 +24,29 @@ var StyleLayer = L.TileLayer.extend({
     _getAttribution: function(_) {
         var styleURL = format_url.style(_, this.options && this.options.accessToken);
         request(styleURL, L.bind(function(err, style) {
-            if (err) {
+            if (err || !style) {
                 util.log('could not load Mapbox style at ' + styleURL);
                 this.fire('error', {error: err});
-            }
-            var sources = [];
-            for (var id in style.sources) {
-                var source = style.sources[id].url.split('mapbox://')[1];
-                sources.push(source);
-            }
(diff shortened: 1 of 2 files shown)


danielkueffer:
I testet mapbox with this commit in IE9. The error message "Cannot read property 'sources' of null" is not showing up anymore. But i still get the error message "Unspecified error".

bsudekum:
@danielkueffer thanks for the heads up. Do you think you could provide me with your style id that is erroring? Feel free to DM me on twitter `@bobws` if you'd like to keep it private.

danielkueffer:
@bsudekum this is the style id that is erroring "cipjh0vk3005bcnngari1zqzz".

rarkins:
FYI, I see this sometimes even when using the regular mapbox light style (unmodified)

rarkins:
@bsudekum do you think this can be merged? #1160 refers specifically to `TypeError: Cannot read property 'sources' of null` which I think this solves?

rarkins:
Please merge :)

PR #1180, opened 2016-10-24
Increase z-index of marker distance display

tmcw:
Also:
- https://www.mapbox.com/mapbox.js/example/v1.0.0/filtering-marker-clusters/

### DIFF
--- a/docs/_posts/examples/v1.0.0/0100-01-01-marker-distances.html
+++ b/docs/_posts/examples/v1.0.0/0100-01-01-marker-distances.html
@@ -19,6 +19,7 @@ pre.ui-distance {
   font-size:11px;
   line-height:18px;
   border-radius:3px;
+  z-index: 4;
   }
 </style>
 <div id='map'></div>


tmcw:
@tristen @mourner z-index of new leaflet is lookin' different - is this something to be expected that we should fix in examples, or should it be fixed in leaflet/mapbox.js?

tristen:
fwiw the z-index rules for mapbox.js are here: https://github.com/mapbox/mapbox.js/blob/a8cc60b400fa00b9eac0ac80c02170340ecd4b76/theme/style.css#L140-L149

In Leaflet they are [a lot higher](https://github.com/Leaflet/Leaflet/blob/master/dist/leaflet.css#L81-L91) which would generally break a bunch of UI's added to other examples or the implementations of other folks code.

tristen:
Capturing from convos: the [culprit is this line](https://github.com/mapbox/mapbox.js/blob/a8cc60b400fa00b9eac0ac80c02170340ecd4b76/theme/style.css#L143) and I don't see a good reason why this would be set as it's an immediate child to the root container. @mourner can you shed more light in https://github.com/Leaflet/Leaflet/pull/3591 as to why `leaflet-pane` would need to be touched in order for the [new pane feature](http://leafletjs.com/reference-1.0.0.html#map-createpane) to work? Ideally we fix the css and not adjust all the examples + cite this as a caveat to users who have borrowed from code examples for their own code.

Issue #1178, opened 2016-10-21
Update docs and examples for v3.0.0


tmcw:
- [x] Docs updated
- [ ] Find and fix broken examples

TODO:
- [x] imagePath now expects a trailing `/` in Leaflet 1.0 so https://www.mapbox.com/mapbox.js/example/v1.0.0/coffeescript/ is showing 404s
  - https://github.com/mapbox/mapbox.js/pull/1179
- [x] z-index of map changed so marker distance doesn't appear
  - https://github.com/mapbox/mapbox.js/pull/1180
- [x] leaflet markercluster is broken https://www.mapbox.com/mapbox.js/example/v1.0.0/leaflet-markercluster/
- [x] https://www.mapbox.com/mapbox.js/example/v1.0.0/stylelayer/ is broken

v3.0.0-beta is deployed to the CDN so that we can test it with the docs site. We should test, deploy, and blog on Monday.

geografa:
@tmcw I'm getting 404s on the API docs at the moment. It's running locally and everything passes with one pending test. 2.4.0 docs seem to be working as well. Could this be a publisher thing?

cc @lyzidiamond @colleenmcginnis

tmcw:
Docs are fixed, the generation script changed to accomodate for Jekyll 3, but ended up writing `.html` files instead of directories with `index.html`s in them.

tmcw:
Closing: docs are now fixed (again), open new issues for any issues that crop up.

PR #1182, opened 2016-10-25
Fix Leaflet v1 requiring you to offset zooms for yourself

tmcw:
Refs #1178

### DIFF
--- a/src/style_layer.js
+++ b/src/style_layer.js
@@ -11,12 +11,14 @@ var StyleLayer = L.TileLayer.extend({
     },
 
     initialize: function(_, options) {
-        L.TileLayer.prototype.initialize.call(this, undefined, options);
 
-        this.options.tiles = this._formatTileURL(_);
-        this.options.tileSize = 512;
-        this.options.zoomOffset = -1;
-        this.options.tms = false;
+        options = options || {};
+        options.tiles = this._formatTileURL(_);
+        options.tileSize = 512;
+        options.zoomOffset = -1;
+        options.tms = false;
+
+        L.TileLayer.prototype.initialize.call(this, undefined, options);
 
         this._getAttribution(_);
     },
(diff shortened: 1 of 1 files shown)


bsudekum:
This fails for zoom level 0 which requests tile `-1/0/0`.

tmcw:
Do you know what the previous behavior was?

bsudekum:
@tmcw it'd load tile `0/0/0`, here is an example: http://bl.ocks.org/bsudekum/0b630fab6ce551784a4c91815a9b1b0a

bsudekum:
One idea here is that in leaflet 0.7, tile 0/0/0 was scaled down to 256px. This is not the case anymore in 1.0.

tmcw:
@mourner I'm not seeing a way through this problem in Leaflet v1.0.1, could you take a look? We need the behavior to be:

| Leaflet Zoom | Tile | Tile Size | Pixel Size on Screen |
| --- | --- | --- | --- |
| 0 | 0/0/0 | 512x512px | 256x256px |
| 1 | 0/0/0 | 512x512px | 512x512px |
| 2 | 1/0/0 | 512x512px | 512x512px |

And so on. We were, remarkably, able to make this work with default Leaflet 0.7 behavior, but I'm finding no combination of parameters that will underscale a tile at one specific zoom in this way in Leaflet v1.0.1. Unfortunately this is a breaking issue for Mapbox.js and very high-impact, since styleLayer is the recommended layer and 512px is the natural size.

cc @bsudekum @yhahn since this is part of 5" quest.

IvanSanchez:
I think it might be time to reopen https://github.com/Leaflet/Leaflet/issues/3165. Implementing that would solve this specific behaviour nicely.

bb-juliogarcia:
@IvanSanchez - I've gone ahead and submitted a pull request for the issue at https://github.com/Leaflet/Leaflet/issues/3165. It is unit tested and working for us on our client

tmcw:
Tracking in issue instead of PR: #1192

PR #1189, opened 2016-11-15
Try to use the minNativeZoom option

tmcw:
@mourner - would this be the intended way to use minNativeZoom? If you run make and then try out `http://localhost:3000/test/manual/style_layer.html`, you'll see this does not seem to do the trick.


tmcw:
Closing in favor of #1192

PR #1194, opened 2016-11-21
Upgrade to Leaflet 1.0.2 and fix style layer tiling

### LINKED ISSUE #1192: Fix Leaflet v1 requiring you to offset zooms yourself
Originally #1182, moving out of PR because I don't know how it'll end up.

---

@mourner I'm not seeing a way through this problem in Leaflet v1.0.1, could you take a look? We need the behavior to be:

| Leaflet Zoom | Tile | Tile Size | Pixel Size on Screen |
|---|---|---|---|
| 0 | 0/0/0 | 512x512px | 256x256px |
| 1 | 0/0/0 | 512x512px | 512x512px |
| 2 | 1/0/0 | 512x512px | 512x512px |

And so on. We were, remarkably, able to make this work with default Leaflet 0.7 behavior, but I'm finding no combination of parameters that will underscale a tile at one specific zoom in this way in Leaflet v1.0.1. Unfortunately this is a breaking issue for Mapbox.js and very high-impact, since styleLayer is the recommended layer and 512px is the natural size.

mourner:
Closes #1192, a successor to #1182. Seems to work properly, but I haven't tested extensively yet — just the style layer manual page so far.

cc @tmcw @bsudekum

### DIFF
--- a/src/style_layer.js
+++ b/src/style_layer.js
@@ -11,13 +11,13 @@ var StyleLayer = L.TileLayer.extend({
     },
 
     initialize: function(_, options) {
-        L.TileLayer.prototype.initialize.call(this, undefined, options);
-
-        this.options.tiles = this._formatTileURL(_);
-        this.options.tileSize = 512;
-        this.options.zoomOffset = -1;
-        this.options.tms = false;
-
+        L.TileLayer.prototype.initialize.call(this, undefined, L.extend({}, options, {
+            tileSize: 512,
+            zoomOffset: -1,
+            minNativeZoom: 1,
+            tms: false
+        }));
+        this._url = this._formatTileURL(_);
         this._getAttribution(_);
     },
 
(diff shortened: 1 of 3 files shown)


mourner:
For some reason, while tests pass for me locally, there are 2 failing tests on Travis. Not sure what's going on.

bsudekum (approved):
Tested locally, looks good to me

bsudekum:
Checking out test failures.

bsudekum:
The tests are failing because [`getTile`](https://github.com/mapbox/mapbox.js/pull/1194/files#diff-e1b538b03ec4f0f4fdced9f144bf186bL71) was removed but it's being called in the tests https://github.com/mapbox/mapbox.js/blob/mb-pages/test/spec/style_layer.js#L95

tmcw:
Behavior is still incorrect:

At zoom 1:

2016-11-21 at 11 55 am

These tiles are from



and downscaled, while the map should use



at its native scale.

tmcw (changes requested):
Exhibits tile downscaling at zoom=1

src/style_layer.js
```diff
@@ -11,13 +11,13 @@ var StyleLayer = L.TileLayer.extend({
...
-
-        this.options.tiles = this._formatTileURL(_);
-        this.options.tileSize = 512;
-        this.options.zoomOffset = -1;
-        this.options.tms = false;
-
+        L.TileLayer.prototype.initialize.call(this, undefined, L.extend({}, options, {
+            tileSize: 512,
+            zoomOffset: -1,
+            minNativeZoom: 1,
```
tmcw: setting `minNativeZoom: 0` fixes the problem I'm seeing.

mourner: Yep, just figured out the same — committed.

tmcw (approved):

tmcw:
👍  Thanks for digging into this Vladimir!

PR #1196, opened 2016-11-25
handle tile url with no format in tile.json files

calvinmetcalf:
if one of the urls in a tile.json tiles array doesn't have a format suffix then mapbox throws an error when it tries to load the tile.json


tmcw:
👍 Thanks, sorry for the delay!

Issue #1198, opened 2016-12-15
geocoder.queryURL does not account for bbox option


lukewhyte:
On line 50 of `src/geocoder.js`, inside of the `geocoder.queryURL` method, each of the possible options to be fed into the query string are handled. `bbox`, despite being a possible option that the API does understand, isn't accounted for here and won't be passed to the query.

lukewhyte:
Also, [API documentation](https://www.mapbox.com/api-documentation/?language=JavaScript#request-format) isn't clear as to whether the `bbox` option should be passed in as an array or converted to a string beforehand.

tmcw:
👍 Fixed in #1199

The REST API documentation is about the URL structure, and urls are strings alone.

PR #1199, opened 2016-12-15
Handle bbox option in geocode.js and add relevant test

lukewhyte:
This is relevant to issue #1198. On line 50 of src/geocoder.js, inside of the geocoder.queryURL method, each of the possible options to be fed into the query string are handled. bbox, despite being a possible option that the API does understand, wasn't accounted for here and, in turn, not passed to the query. This PR fixes that.

Test has been added too.

### DIFF
--- a/src/geocoder.js
+++ b/src/geocoder.js
@@ -64,6 +64,14 @@ module.exports = function(url, options) {
                 }
             }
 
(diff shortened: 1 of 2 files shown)


tmcw:
👍 Awesome, thanks Luke!

PR #1224, opened 2017-05-04
Run make to create built release before NPM publish

andi1984:
At the moment each project depending on mapbox.js has to create the built release files on their own. With that we bring the built releases directly to these projects by putting them inside the package before publish, so they are available for everyone directly when the package is installed via npm.

The git repository instead is still clean and will not contain them.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+node_modules
+_site
+tmp
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "sanitize-caja": "0.1.4"
   },
   "scripts": {
-    "test": "eslint --no-eslintrc -c .eslintrc src && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html"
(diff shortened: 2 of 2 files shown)


jfirebaugh:
Can you explain the rationale here? If you're installing Mapbox.js via npm, our expectation is that you want to use CommonJS and the package `main` at `src/index.js`, rather than bundled and minified dist files.

andi1984:
Hello @jfirebaugh,

NPM is getting more and more being **the one** package manager. Whether it is for node or frontend related components.

I can partly understand your assumption, but imho it is pretty common in our industry to deliver source and distribution ready files when fetched from a package manager like npm or bower. So I don't see a valid reason just to provide this functionality in form of an extra repository for bower users only. And it doesn't harm you in any way as the git repo can stay the same, you just deploy the distribution files to npm also.

Please take some time to think about it. I really don't see any downsides of providing the distribution files also via npm and not just via bower. So you leave it open to your supporters and users which package manager they are using and thus they are more independent.

Greets,

Andi

PR #1225, opened 2017-05-23
Replace instances of `map-feedback` with `feedback`

tristen:


jfirebaugh:
Does `/feedback/` use the same zoom level conventions as `/map-feedback/` (i.e. z0 = world in 256 pixels), or does it use GL zoom level conventions (i.e. world in 512 pixels)? If the latter, doesn't Mapbox.js need to apply a zoom level adjustment?

tristen:
mapbox.js passes a map.id as the first argument in the hash. If that's detected, A user will see a map rendered in mapbox.js and not Mapbox GL JS.

jfirebaugh (approved):

Issue #1250, opened 2018-01-10
maxNativeZoom on TileLayer


tvblomberg:
Hello,

I am currently in the process of upgrading from version 2.1.5 to 3.1.1.
I ran into an issue with the maxNativeZoom option on tile layer.

Basically it looks like it is being ignored and the map keeps zooming displaying white tiles.
My expectation is the tiles should be autoscaled.

This worked find in version 2.1.5. Here are the fiddles:
2.1.5: https://jsfiddle.net/jfwkq0s4/4/
3.1.1: http://jsfiddle.net/dbpfcoqo/6/

Simple code example:
`
L.mapbox.accessToken = 'pk.eyJ1IjoidHZibG9tYmVyZyIsImEiOiJjaXZsZmpsMWwwNXBuMnRudmJqNGQyMjJwIn0.842ovgKydac51tg6b9qKbg';
var map = L.mapbox.map('map', null, { maxZoom: 20})
    .setView([40, -74.50], 9);
    
var layer = new L.mapbox.tileLayer('mapbox.streets', { maxNativeZoom: 17, maxZoom: 20 });

map.addLayer(layer); 
`

tvblomberg:
See my solution here:
https://stackoverflow.com/questions/48195486/is-there-an-issue-with-maxnativezoom-on-tile-layers-in-mapboxjs/48209306#48209306

Not sure if this is the best approach curious on anyone's thoughts.

tvblomberg:
I will make a pull request for this tonight when I have time.

PR #1251, opened 2018-01-12
Fix Overzoom issues with tile layers

tvblomberg:
This pull request should fix issues with calculating zoom for maxNativeZoom and calculating tilesize.
Right now the maxNativeZoom option does not do anything for tileLayers.  This worked correctly in version 2.1.5. Currently users will get white tiles when overzooming. This is my first pull request so please let me know if I am doing something wrong. This should fix issue #1250 

Here is the example of the two versions:
2.1.5: https://jsfiddle.net/jfwkq0s4/4/
3.1.1: http://jsfiddle.net/dbpfcoqo/6/

### DIFF
--- a/src/tile_layer.js
+++ b/src/tile_layer.js
@@ -52,11 +52,14 @@ var TileLayer = L.TileLayer.extend({
           }
         }
 
+        var minZoom = this.options.minZoom || json.minzoom || 0;
+		var maxZoom = this.options.maxZoom || json.maxzoom || 18;
+
         L.extend(this.options, {
             tiles: json.tiles,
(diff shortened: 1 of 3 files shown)


olaf-cloudgenix:
I hit this too when setting detectRetina to true on the tileLayer.  It tries to fetch either tiles for the completely wrong geography, or tiles that don't exist (resulting in 404 errors).  This pull request fixes it.

riastrad:
I'd like to take this along with the next minor release, but will be using https://github.com/mapbox/mapbox.js/pull/1323 to track adding it against latest master instead of wrestling with a +2 year rebase.

PR #1255, opened 2018-04-05
Remove broken Foursquare example

colleenmcginnis:
This PR removes the broken Foursquare example, and redirects that URL to the main examples page. 

cc @alulsh


alulsh (approved):
LGTM @colleenmcginnis! Thanks for putting this together.

PR #1258, opened 2018-05-10
support limit parameter in geocoder

tomap:
geocoder api supports between 1 and 10 search results. Default value is 5.
Now, it's possible to set the limit


tomap:
Hey, you published a new version recently, could you look at the PR?

riastrad:
Thanks for flagging this, @tomap. As far as I can tell this looks like a pretty straightforward change, but before we approve/merge would you mind doing the following two things:

- [ ] add a unit test in https://github.com/mapbox/mapbox.js/blob/publisher-production/test/spec/geocoder.js for this new query parameter
- [ ] update the CHANGELOG.md to reflect this addition

tomap:
4 years later, would you mind doing it? I did half the job. I would be very glad if you did the rest of the job :)

PR #1268, opened 2018-11-01
Set up for publisher v2

danswick:
Adds an `.artifacts.yml` to set up for publisher v2, also enabling @jseppi to update this repo's page shell.

cc @colleenmcginnis


jseppi (approved):
Thanks, @danswick!

PR #1269, opened 2018-11-10
[WIP] Encourage mapbox.js users to use GL styles

HeyStenson:
This PR updates Mapbox.js examples so that they use GL styles instead of Mapbox core classic styles. 

- **11/9:** Updated 85/153 examples. These are straight swaps of GL styles instead of classic styles, using `.addLayer(L.mapbox.styleLayer('style'));`. 
- **Next steps:** Next week I'll tackle the examples that are trickier - they use custom tilesets, involve more complex .js functions, etc.

cc @brsbl @ian29 @colleenmcginnis


HeyStenson:
@colleenmcginnis Finally finished all the examples! 💃 I also bumped the style version numbers when applicable. Ready for you to take another look when the time is right.

PR #1273, opened 2019-01-08
Update example.html

geeknik:
s/http/https/g

This patch was submitted to possibly fix a mixed content issue that I reported on [HackerOne](https://hackerone.com/reports/476641).


geeknik:
I was hoping to get some [HackerOne rep points](https://hackerone.com/reports/476641) for the patch but it was closed by HackerOne staff before y'all could see it.

oliikit (approved):
🎉 thanks for taking proactive measures here, @geeknik!
screen shot 2019-01-10 at 10 33 31 am

PR #1275, opened 2019-01-09
Change feedback app URL

davidtheclark:
We are transitioning the feedback app to a new subdomain: `apps.mapbox.com/feedback/`.

### DIFF
--- a/docs/_posts/examples/v1.0.0/0100-01-01-leaflet-studio-style.html
+++ b/docs/_posts/examples/v1.0.0/0100-01-01-leaflet-studio-style.html
@@ -19,6 +19,6 @@ L.tileLayer(
     'https://api.mapbox.com/styles/v1/mapbox/emerald-v8/tiles/{z}/{x}/{y}?access_token=' + L.mapbox.accessToken, {
         tileSize: 512,
         zoomOffset: -1,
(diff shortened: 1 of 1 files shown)


davidtheclark:
Hm, there are some other changes that could be made to this repo. Do you think it's worth doing, or should we just leave this as-is and let redirects do their thing? cc @mapbox/docs 

I also don't know who to ping about changes to the code itself. @agentpu1011 what do you think?

colleenmcginnis:
>Hm, there are some other changes that could be made to this repo. Do you think it's worth doing, or should we just leave this as-is and let redirects do their thing?

What kind of changes @davidtheclark? 

>I also don't know who to ping about changes to the code itself.

cc @brsbl @chloekraw 😬

davidtheclark:
@colleenmcginnis one more doc change (https://github.com/mapbox/mapbox.js/blob/62b2db755a0045001589c1336eeb6f75926a28ca/docs/_posts/examples/v1.0.0/0100-01-01-plain-leaflet.html#L15) but also a change to the library: https://github.com/mapbox/mapbox.js/blob/62b2db755a0045001589c1336eeb6f75926a28ca/test/helper.js#L46

colleenmcginnis (approved):
🍍

PR #1279, opened 2019-02-08
[docs] adds jekyll-sitemap plugin to generate a sitemap

katydecorah:
We're adding a sitemap to Mapbox.js so we can add it to the search console.

To test this PR:

1. `bundle install`
2. `bundle exec jekyll serve`
3. Navigate to http://127.0.0.1:4000/mapbox.js/sitemap.xml  and you should see an XML document with the files in this site!


colleenmcginnis (approved):
🎸

PR #1280, opened 2019-02-09
sitemap, take 2

katydecorah:
@colleenmcginnis apologies for the double PR. Looks like #1279 did not take because I didn't added the plugin to the other config files (so it worked locally, but not in production). Also, I think we need to commit the Gemfile.lock after all so publisher knows to install the plugin. 

I pushed this branch to staging in case you'd like to double check that /mapbox.js/sitemap.xml exists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ dist
 node_modules
 _site
 tmp
-Gemfile.lock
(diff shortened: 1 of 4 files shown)


colleenmcginnis (approved):
👌

PR #1294, opened 2019-06-19
[docs] remove search bar

katydecorah:
The search bar on this site hasn't been working properly for sometime. Now that we've launched docs-wide search, I think we should remove this particular search bar. (I don't think using our new search engine here is an option because we filter out Mapbox.js API docs.)

I pushed these changes to staging for faster review.


colleenmcginnis (approved):
💐

PR #1296, opened 2019-07-11
[DNM] Change "map ID" to "tileset ID"

HeyStenson:
The term map ID is a holdover from when Mapbox map styles and map tiles were a unified concept (i.e. Mapbox Studio Classic styles). Now a map ID actually refers to a tileset ID (and the map ID is often confused for a style ID). Changing the term _map ID_ to _tileset ID_ will clarify these concepts. 

This PR makes that change for the docs for v3.2.0. (Due to an early misunderstanding of how to update, it also makes this change to v1.6.0 😅) Let me know if you think this change should be made to all versions, not just the latest. 

**Timing:** This PR should be merged when related PRs are merged, which will likely be 7/17.


HeyStenson:
closing in favor of https://github.com/mapbox/mapbox.js/pull/1297

PR #1299, opened 2019-10-04
create redirect for leaflet-image plugin

katydecorah:
This PR creates a redirect for /mapbox.js/example/v1.0.0/leaflet-image/ (back to main examples page) and removes the "View example" from the plugins page for the Leaflet Image plugin:

image

Fixes https://github.com/mapbox/subdomain-docs/issues/49

🐌


HeyStenson (approved):
Looks good to me! Thanks for doing this @katydecorah!

PR #1305, opened 2019-10-16
Update src/format_url.js: fixes #1304

### LINKED ISSUE #1304: Don't force HTTPS on style URL's
If `HTTP_URL` is set and `FORCE_HTTPS` is false all requests use HTTP with the exception of the first style fetch request.

The root cause is in the  `style` function in the `format_url` module forces HTTPS regardless of the config. 

https://github.com/mapbox/mapbox.js/blob/a606ef3d082a55a48983a2197822b5446babed09/src/format_url.js#L52-L60

The `request` module is already equipped with making decisions on the protocol based on the config and other variables, why not defer that decision to the request module?

I propose we remove the explicit replace in line https://github.com/mapbox/mapbox.js/blob/publisher-production/src/format_url.js#L57

whyvez:
Fixes issue where style URLs where always being requested from HTTPS even if the config was set to use HTTP.

### DIFF
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -53,8 +53,7 @@ module.exports.style = function(styleURL, accessToken) {
     if (styleURL.indexOf('mapbox://styles/') === -1) throw new Error('Incorrectly formatted Mapbox style at ' + styleURL);
 
     var ownerIDStyle = styleURL.split('mapbox://styles/')[1];
-    var url = module.exports('/styles/v1/' + ownerIDStyle, accessToken)
-        .replace('http://', 'https://');
+    var url = module.exports('/styles/v1/' + ownerIDStyle, accessToken);
 
     return url;
 };


asheemmamoowala (commented):
In #1285 `FORCE_HTTPS` was set to `true` by default. Is this change needed even after setting `FORCE_HTTPS: false`? If so, maybe the appropriate fix is to have the `style` method respect the `FORCE_HTTPS` configuration option

whyvez:
@asheemmamoowala The `FORCE_HTTPS` config is already used from the `request` module. That by itself should be sufficient to detect and apply the proper protocol to the style requests.

asheemmamoowala (approved):
Given that `FORCE_HTTPS` will continue to work correctly after this change, Im 👍 with making it possible to respect existing `http` protocol

PR #1308, opened 2019-10-17
Add release.sh: automated releases!

whyvez:
* Add an automated release script
* Renamed `DEPLOYMENT.md` to `RELEASE.md`
* Update `RELEASE.md`


RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+### Manual Deployement
+
+* Update `_config.yaml`, `_config.publisher-production.yml`, `_config.publisher-staging.yml`
+* Build docs. `./deploy.sh v<major.minor.patch>`
+* Commit docs.
+* Bump version and tag. `npm version <major.minor.patch>`
+* Push to Github. `git push origin publisher-production --tags`
+* Build `mapbox.js` `make`
+* Publish to CDN. `./deploy.sh v<major.minor.patch>`
+* Publish to NPM. `mbx npm publish`
```
alulsh: @whyvez I had to run regular `npm publish` while logged in as my personal npm account (https://www.npmjs.com/~alulsh). The mbx wrapper does not work here. Anybody who is an owner of this package (`npm owner ls mapbox.js`) can run `npm publish` here.

release.sh
```diff
@@ -0,0 +1,54 @@
...
+if [[ "$should_publish_cdn" == "y" ]]; then
+  make
+  ./deploy.sh "$tag"
+fi
+
+echo "Do you want to publish NPM? (y/n)"
+read -r should_publish_npm
+
+if [[ "$should_publish_npm" == "y" ]]; then
+  mbx npm publish
```
alulsh: @whyvez this should also be `npm publish` without the mbx wrapper. We should probably do some light error handling early on in the script. We should check that people have ownership rights on the mapbox.js npm module (`npm owner ls mapbox.js`) before the rest of the script can run.

RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+  issues in question
+
+### Automated Deployement
+
+`npm run release <major.minor.patch>`
+
+### Manual Deployement
+
+* Update `_config.yaml`, `_config.publisher-production.yml`, `_config.publisher-staging.yml`
+* Build docs. `./deploy.sh v<major.minor.patch>`
```
alulsh: @whyvez the mapbox.js build should be published and working before docs are updated, right? Otherwise, the documentation will fail to load the latest mapbox.js release. If this is the case we should move the documentation update steps to its own separate section at the end. Looping in @katydecorah for guidance here.

whyvez: @alulsh I believe the actual issue was that the version in the `_config*.yml` files were bumped but the documentation pages were not generated. You can see this in the test added test here: https://github.com/mapbox/mapbox.js/pull/1309. I've also updated the script to make the config version and docs update in the same commit.

RELEASE.md
whyvez: Updated.

release.sh
whyvez: Updated.

RELEASE.md
danswick: Your explanation looks right to me, @whyvez. Updating the version number in `_config.yml` will cause the mapbox.js site to attempt to redirect to an `index.html` within a directly with the same name as the version number. So, if you tried to go to: 

```
https://www.mapbox.com/mapbox.js/
```

You would be redirected to 

```
https://docs.mapbox.com/mapbox.js/api/v3.2.1/
```

An ideal update would both 1) bump the version number and 2) include newly-generated docs for that specific version.

whyvez: Updated.

danswick (commented):
@whyvez, these changes look good from a docs generation and deployment perspective, but I would defer to @asheemmamoowala or someone else from the core Mapbox GL JS team since they're ultimately responsible for the direction of this repo!

PR #1309, opened 2019-10-17
assert documentation exists when mapboxjs version is bumped

katydecorah:
When `mapboxjs` is updated in _config.publisher-production.yml the documentation for that version must exist, otherwise, the production site will redirect to the 404 page. This is because the Mapbox.js landing page redirects to the latest version of the documentation.

This PR adds a test to verify that the documentation folder exists for the current `mapboxjs` version in _config.publisher-production.yml. If the folder does not exist, the tests will fail. 

cc @whyvez @alulsh

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "sanitize-caja": "0.1.4"
   },
   "scripts": {
-    "test": "eslint src && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html",
+    "test": "eslint src && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html && mocha test/docs.js",
     "prepublishOnly": "npm run build",
     "build": "make"
   },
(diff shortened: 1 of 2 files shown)


whyvez:
@katydecorah When will this test run? I don't think tests are run in Travis. Running tests locally is also hard because of the required dependencies (i.e. phantomjs). I think tests could fail and nobody would notice.

danswick:
@whyvez the test was run on this PR https://travis-ci.org/mapbox/mapbox.js/builds/599273048#L520. I think the idea here is that docs should be created _first_, then version numbers bumped.

whyvez:
@danswick Right! I forgot about the implicit `npm test` in Travis. When you say version bumped, you mean the version in the `_config*.yml`? If so why? I created an automated script [here](https://github.com/mapbox/mapbox.js/pull/1308) and would like to get it right in there.

katydecorah:
The automated script in #1308 is awesome. I would still like to include this test as added assurance that documentation exists when the version is bumped in the publisher config to prevent broken pages - this is critical to documentation's operational excellence.

danswick (approved):
This test looks great, @katydecorah!

PR #1317, opened 2020-01-29
update default tile/style endpoints for mapbox styles

### LINKED ISSUE #1316: Point to modern Static Tiles API by default
A large amount of Mapbox.js's current defaults for `L.mapbox.map` point to a Legacy Static Tile API. These defaults should be updated to point to the [Mapbox Static Tiles API](https://docs.mapbox.com/api/maps/#static-tiles) instead.

riastrad:
Resolves #1316 

Up until now, Mapbox.js defaults have relied heavily on the Mapbox Legacy Static Tiles endpoints.

I'll mark this PR as ready for review when the following are implemented:
- [x] Updates initialization code so that passing in a template `mapID` will automatically load the equivalent style from the modern Static Tiles API.
- [x] Set console warning when automatic support for classic style IDs (e.g. `riastrad.{class style id}`) is passed to `map` objects and `tileLayer` objects.
- [x] ~Logs a console warning about dropping support for the outdated `tileLayer.setFormat()` method with default `mapIDs`.~  Given that the url pattern for legacy & modern APIs is different, this appears to be unnecessary.
- [x] All previous tests are updated and passing
- [x] new tests for changes to default behavior

If I identify any other usage of legacy APIs within the defaults, I will flag it and incorporate it within this PR before the review.

cc @mapbox/static-apis

### DIFF
--- a/src/tile_layer.js
+++ b/src/tile_layer.js
@@ -79,7 +79,14 @@ var TileLayer = L.TileLayer.extend({
 
         var templated = L.Util.template(url, tilePoint);
         if (!templated || !this.options.format) {
-            return templated;
+            var tileURL;
+            if (L.Browser.retina && url.indexOf('/styles/v1') !== -1) {
+                tileURL = [templated.slice(0, templated.indexOf('?')), '@2x', templated.slice(templated.indexOf('?'))].join('');
+            } else {
+                tileURL = templated;
+            }
+
+            return tileURL
         } else {
             return templated.replace(formatPattern,
                 (L.Browser.retina ? this.scalePrefix : '.') + this.options.format);
--- a/src/tile_layer.js
+++ b/src/tile_layer.js
@@ -81,7 +81,7 @@ var TileLayer = L.TileLayer.extend({
         if (!templated || !this.options.format) {
             var tileURL;
             if (L.Browser.retina && url.indexOf('/styles/v1') !== -1) {
-                tileURL = [templated.slice(0, templated.indexOf('?')), '@2x', templated.slice(templated.indexOf('?'))].join('');
+                tileURL = templated.replace('?', '@2x?');
             } else {
                 tileURL = templated;
             }
--- a/src/util.js
+++ b/src/util.js
@@ -8,6 +8,8 @@ function contains(item, list) {
     return false;
 }
 
+var warnOnceHistory = {};
+
 module.exports = {
     idUrl: function(_, t) {
         if (_.indexOf('/') === -1) t.loadID(_);
@@ -21,8 +23,11 @@ module.exports = {
         }
     },
     warn: function(_) {
+        // avoid cluttering the console with duplicative warnings
+        if (warnOnceHistory[_]) return;
         if (typeof console === 'object' && 
             typeof console.warn === 'function') {
+            warnOnceHistory[_] = true;
             console.warn(_);
         }
         /* eslint-enable no-console */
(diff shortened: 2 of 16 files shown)


riastrad:
👋 @mapbox/gl-js @mapbox/static-apis this branch changes some pretty core functionality, so would appreciate a close review from one or both of you.

test/spec/map.js
```diff
@@ -53,11 +53,25 @@ describe('L.mapbox.map', function() {
                 done();
             });
 
-            server.respondWith("GET", "https://a.tiles.mapbox.com/v4/mapbox.map-0l53fhk2.json?access_token=key",
+            server.respondWith("GET", "https://api.mapbox.com/v4/mapbox.map-0l53fhk2.json?access_token=key&secure",
                 [200, { "Content-Type": "application/json" }, JSON.stringify(helpers.tileJSON)]);
             server.respond();
         });
 
+        it('loads styleJSON from a default mapbox ID', function(done) {
```
rmrice, line 61: Can we add a test like this for more GL styles?

riastrad: I hesitated to do this, since there are 7 template styles and the functionality to intercept them & redirect to the styleJSON endpoint is the same for each:

https://github.com/mapbox/mapbox.js/blob/f9990352a3b4edeca4edca6e7cd035ffa4fd60c2/src/config.js#L8-L17

If you don't feel like it would clutter the test suite with too many redundant tests, then happy to go ahead and add 6 more. 👍

rmrice: Gotcha. Then I think just one test is fine.

riastrad:
👋 @asheemmamoowala @mourner any chance you'll be able to review this before the end of the week?

mourner:
@riastrad yes, will do!

src/tile_layer.js
```diff
@@ -79,7 +79,14 @@ var TileLayer = L.TileLayer.extend({
 
         var templated = L.Util.template(url, tilePoint);
         if (!templated || !this.options.format) {
-            return templated;
+            var tileURL;
+            if (L.Browser.retina && url.indexOf('/styles/v1') !== -1) {
+                tileURL = [templated.slice(0, templated.indexOf('?')), '@2x', templated.slice(templated.indexOf('?'))].join('');
```
mourner: Nit: this could probably be written in a simpler way, maybe `templated.replace('?', '@2x?')`?

src/util.js
```diff
@@ -19,6 +19,12 @@ module.exports = {
             typeof console.error === 'function') {
             console.error(_);
         }
+    },
+    warn: function(_) {
+        if (typeof console === 'object' && 
+            typeof console.warn === 'function') {
+            console.warn(_);
+        }
```
mourner, line 32: The context this is being used may cause hundreds of duplicate warnings to be printed — maybe worth adopting something like `warnOnce` [from GL JS](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/util.js#L318-L332)?

mourner (approved):
👍 looks great to me apart from a few nits.

src/util.js
riastrad: Great point, implemented in https://github.com/mapbox/mapbox.js/pull/1317/commits/5a25f6241c21be672d67d2859bdfe0aeec3f0b0d

src/tile_layer.js
riastrad: Wow, yep - would make this much more readable. Switched to `.replace` in https://github.com/mapbox/mapbox.js/pull/1317/commits/bd3b7bc38427863e16463b8e1f8bde65144d0c90

mourner (approved):

PR #1322, opened 2020-03-04
[docs] correct api usage of plain-leaflet example

### LINKED ISSUE #1320: [docs] incorrect implementation in the leaflet example
The Leaflet.js example that is live here: https://docs.mapbox.com/mapbox.js/example/v1.0.0/plain-leaflet/

Does not correctly account for the `tileSize` of the responses coming from the [Static Tiles API](https://docs.mapbox.com/api/maps/#static-tiles). The fix should be quite simple:

```
var mapboxTiles = L.tileLayer('https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/{z}/{x}/{y}?access_token=' + L.mapbox.accessToken, {
       attribution: '© <a href="https://www.mapbox.com/feedback/">Mapbox</a> © <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>',
       tileSize: 512,
       zoomOffset: -1
});
```

riastrad:
Fixes #1320 

Previously, our "plain leaflet" example requested tiles from the Static Tiles API without account for the fact that this API returns 512x512 styles by default, whereas Leaflet.js expects 256x256 tiles by default. This update corrects this example to account for this discrepancy.

(I also updated `DOCUMENTING.md` while I was in there.)

cc/ @mapbox/static-apis @mapbox/docs

### DIFF
--- a/DOCUMENTING.md
+++ b/DOCUMENTING.md
@@ -27,7 +27,7 @@ make
 jekyll serve -w -p 4000
 ```
 
-Edit the site. Existing examples [are located here](https://github.com/mapbox/mapbox.js/tree/mb-pages/docs/_posts/examples/v1.0.0) -
+Edit the site. Existing examples [are located here](https://github.com/mapbox/mapbox.js/tree/publisher-production/docs/_posts/examples/v1.0.0) -
 copy one to a new file with the same naming convention to start a new example. Test your new example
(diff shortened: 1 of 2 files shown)


katydecorah (approved):
Thanks @riastrad!

PR #1323, opened 2020-03-18
correct overzooming bug when maxZoom options are specified by user

### LINKED ISSUE #1250: maxNativeZoom on TileLayer

riastrad:
Resolves #1250 

Back in 2018, @tvblomberg put up a PR that solved an overzooming bug (ref. https://github.com/mapbox/mapbox.js/pull/1251). 

The bug fix is sound, but enough time has passed that I think pulling these changes into a new PR and testing them against the current state of the repo. I also want to add a manual test that can be used to quickly determine if this has regressed.

### DIFF
--- a/src/tile_layer.js
+++ b/src/tile_layer.js
@@ -52,11 +52,14 @@ var TileLayer = L.TileLayer.extend({
           }
         }
 
+        var minZoom = this.options.minZoom !== 0 ? this.options.minZoom : json.minzoom || 0;
+        var maxZoom = this.options.maxZoom !== 18 ? this.options.maxZoom : json.maxzoom || 18;
+
         L.extend(this.options, {
             tiles: json.tiles,
             attribution: this.options.sanitizer(json.attribution),
-            minZoom: json.minzoom || 0,
-            maxZoom: json.maxzoom || 18,
+            minZoom: minZoom,
+            maxZoom: maxZoom,
             tms: json.scheme === 'tms',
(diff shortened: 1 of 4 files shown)


riastrad:
Checking with the manual test, the patch does not appear to resolve the erroneous behavior. 

Kapture 2020-03-18 at 16 29 28

So this one actually requires more digging, and as a result I'm going to descope it from the v3.3.0 release.

PR #1324, opened 2020-03-18
fix bug where logo did not appear if layer added after map init

### LINKED ISSUE #1314: Fix for missing wordmark attribution on older mapbox.js maps
We were having some issues with missing attribution on maps built with a mapbox.js widget made by Action Network. And it looks like watermark attribution is missing from at least some of our older examples as well, like: https://docs.mapbox.com/mapbox.js/example/v1.0.0/ 

Jason from Action Network was able to figure out a solution :tada: and I thought I'd share it here in case anyone else has run into this:

> I was able to get the logo to show up by setting a map ID per the second option here: https://docs.mapbox.com/mapbox.js/api/v3.2.1/l-mapbox-map/
>
> So
>
`map = L.mapbox.map('map_canvas');`
>
> became
>
`map = L.mapbox.map('map_canvas','mapbox.streets');`
>
> Seems like you need to set that map ID or the logo won't show up, even though that's not a required option.

riastrad:
Fixes https://github.com/mapbox/mapbox.js/issues/1314, where the mapbox logo would not be added to maps when they added the mapbox tilelayer's after initialization of `L.mapbox.map`. 

However, note that it does not correct the behavior noted in https://github.com/mapbox/mapbox.js/issues/1227, as there are still problems with the logo appearing when a `styleLayer` is added to the map after the map object's initialization. 

cc/ @mapbox/static-apis

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -209,7 +209,7 @@ var LMap = L.Map.extend({
 
         // ensure logo appears even when mapbox layer added after map is initialized
         var mapboxLogoControl = this._mapboxLogoControl.getContainer();
-        if (!L.DomUtil.hasClass(mapboxLogoControl)) {
+        if (!L.DomUtil.hasClass(mapboxLogoControl, 'mapbox-logo-true')) {
             var tileJSON = layer.getTileJSON();
             this._mapboxLogoControl._setTileJSON(tileJSON);
         }
--- a/src/map.js
+++ b/src/map.js
@@ -209,15 +209,9 @@ var LMap = L.Map.extend({
 
         // ensure logo appears even when mapbox layer added after map is initialized
         var mapboxLogoControl = this._mapboxLogoControl.getContainer();
-        if (
-            (layer.options.tiles) &&
-            (layer.options.tiles[0].indexOf('mapbox.com') !== -1) &&
-            !(L.DomUtil.hasClass(mapboxLogoControl))
-        ) {
+        if (!L.DomUtil.hasClass(mapboxLogoControl)) {
             var tileJSON = layer.getTileJSON();
-            if (!(tileJSON.mapbox_logo) || (tileJSON.mapbox_logo === true)) {
-                this._mapboxLogoControl._setTileJSON(tileJSON);
-            }
+            this._mapboxLogoControl._setTileJSON(tileJSON);
         }
 
         this._updateMapFeedbackLink();
--- a/src/map.js
+++ b/src/map.js
@@ -209,8 +209,15 @@ var LMap = L.Map.extend({
 
         // ensure logo appears even when mapbox layer added after map is initialized
         var mapboxLogoControl = this._mapboxLogoControl.getContainer();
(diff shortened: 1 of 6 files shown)


src/map.js
```diff
@@ -207,6 +207,13 @@ var LMap = L.Map.extend({
             this._zoomBoundLayers[L.stamp(layer)] = layer;
         }
 
+        // ensure logo appears even when mapbox layer added after map is initialized
+        var mapboxLogoControl = this._mapboxLogoControl.getContainer();
+        if (!L.DomUtil.hasClass(mapboxLogoControl)) {
```
jaericson: `hasClass` expects 2 arguments here, so currently the `if` logic here will always be `true`

src/mapbox_logo.js
```diff
@@ -22,6 +22,11 @@ var MapboxLogoControl = L.Control.extend({
         if (json.mapbox_logo) {
             L.DomUtil.addClass(this._container, 'mapbox-logo-true');
         }
+
+        // account for usage of styleJSON as functional tileJSON for defaults
+        if (!json.tilejson && (json.owner && json.owner === 'mapbox')) {
```
jaericson: I think the `json.owner` check is superfluous; you'd be okay with

`if (!json.tilejson && json.owner === 'mapbox') {`

since we expect `json` to be an object and if there's no `owner` key, it'll just return `undefined` for `json.owner`.

jaericson (changes requested):
Thanks for tackling this!

src/map.js
riastrad: Yep, good catch. Fixed in 36eb4d83

src/mapbox_logo.js
riastrad: Good point. 👍 simplified in 97179796

test/spec/mapbox_logo.js
```diff
@@ -17,11 +17,17 @@ describe('mapbox_logo', function() {
...
 
     it('is not on tilejson map without mapbox_logo flag', function() {
-        var map = L.mapbox.map(element, tileJSON);
+        var map = L.mapbox.map(element, helpers.tileJSON_mapboxlogoMissing);
         var mapboxLogoControl = map._mapboxLogoControl.getContainer();
         expect(L.DomUtil.hasClass(mapboxLogoControl, 'mapbox-logo-true')).to.be(false);
     });
 
+    it('is on map when tilejson is mapbox styleJSON without mapbox_logo flag', function() {
+        var map = L.mapbox.map(element, helpers.styleJSON);
```
jaericson, line 26: `helpers.styleJSON` has `"mapbox_logo": true`, so either the title of this test needs to change or the `mapbox_logo` key needs to be removed from `helpers.styleJSON`.

jaericson (changes requested):
Just one more comment/question

test/spec/mapbox_logo.js
riastrad: oh, blorp - yeah. taken care of in 5aec34ac

jaericson (approved):
👍 thanks Josh!

Issue #1327, opened 2020-04-08
L.mapbox.styleLayer('mapbox://styles/mapbox/streets-v11') throws browser warning about loading "Studio Classic" styles scheduled for deprecation 


george700dl:
Hello,  we are getting our code ready for the 6/1/2020 deprecation of the Classic styles. We use MapboxGL, but we use MapboxJS (v 3.3.0 now) as a fallback for non-webGL supporting browsers. We've been using MapboxJS since 2013 and MapboxGL since last year.

This example itself: https://docs.mapbox.com/mapbox.js/example/v1.0.0/ throws:

1. A warning about loading classic styles: "Warning: this implementation is loading a Mapbox Studio Classic style (mapbox.mapbox-streets-v8,mapbox.mapbox-terrain-v2). Studio Classic styles are scheduled for deprecation"

2. A Leaflet error: Uncaught TypeError: t.getTileJSON is not a function
    at e._updateLayer (map.js:213)
    at e._onLayerReady (map.js:195)
    at e.fire (leaflet-src.js:593)
    at e.<anonymous> (style_layer.js:46)
    at XMLHttpRequest.r (request.js:28)
    at XMLHttpRequest.e (corslite.js:45)

Version prior to 3.3.0 do not throw these. My main concern is the style references. My understanding was that 'mapbox://styles/mapbox/streets-v11' was the way to go forward past 6/1/2020. Why the loading of mapbox.mapbox-streets-v8 and mapbox.mapbox-terrain-v2 ?

Thanks for your help.
George

chelseadelgrande:
I have this issue too when trying to create multiple choropleth maps. The documentation throws https://docs.mapbox.com/mapbox.js/example/v1.0.0/choropleth-joined-data-multiple-variables/ and many of the other examples for mapbox js seem to have the same issue.

Chelsea

riastrad:
This is happening because the when the warning was added in #1317, it failed to account for the fact that `L.mapbox.styleLayer` objects set the tileJson value based on the Mapbox style object's sources field: 

https://github.com/mapbox/mapbox.js/blob/6a6db371daa4ac49940948417ad2c09cf51cdcbb/src/style_layer.js#L32-L48

The good news is that, if you are using `L.mapbox.styeLayer` in your implementation and not directly adding Mapbox classic style as a `L.mapbox.tileLayer`, then these errors can safely be ignored.

chelseadelgrande:
[...]

PR #1328, opened 2020-04-17
Fix choropleth example

malwoodsantoro:
The _**Multiple choropleth maps example**_ is currently broken. After further investigation, it appears that this is happening because `usLayer.getGeoJSON()` is not returning the data in time for it to be used in the joinData function (and is ultimately returning `undefined`). 

choropleth-broken

I tried a few fixes, but wrapping joinData in a setTimeout function seems to give getGeoJSON() the time it needs to be used later in the function. Something strange to me is that calling `usLayer.getGeoJSON()` and setting it to a variable before the setTimeout function still returns `undefined` in joinData. I have to pass the function `usLayer.getGeoJSON()` rather than the variable name for it to work properly. 

choropleth-fix

@katydecorah would you be able to take a look at my proposed changes? I'm wondering if there is a better way to do this without setTimeout(), perhaps using await/async in some way? Would love your thoughts.


katydecorah:
@malwoodsantoro this is interesting 🤔

I believe that this branch is behind publisher-production which is giving us a false positive. I think the issue is that when we updated this example from using a map ID to using a styleLayer, the featureLayer is no longer calling the function `loadData` which kicks off the handling the data.


I created a CodePen to illustrate: https://codepen.io/katydecorah/pen/01a9493b457f97626886b4d689c96ec0?editors=0010

Here's a quick summary of what that CodePen seeks to show:

## Example works when rendering a map id

> image

* `loadData` and `joinData` are logged to the console and the data is loaded properly.

## Example doesn't work with styleLayer

> image

* `loadData` and `joinData` are **not** logged to the console as these functions do not appear to get rendered and so the data is not loaded properly.

---

I'm not very strong on styleLayer and what a solution could look like; @riastrad or @danswick would you happen to have any suggestions?

malwoodsantoro:
Thanks for this, @katydecorah!

> I believe that this branch is behind publisher-production 

Oh, that's my bad. I believe I should be caught up now. 

I now see what you mean. It looks like loadData (and then joinData) is never called because the "ready" event isn't firing: 

```
var usLayer = L.mapbox
  .featureLayer()
  .loadURL('/mapbox.js/assets/data/us.geojson')
  .addTo(map)
  .on("ready", loadData);
```

With `var map = L.mapbox.map("map", "mapbox.light")`: 

Screen Shot 2020-04-20 at 7 03 06 PM

With styleLayer, "ready!" is never logged. 

I realize that, with my fix, I was forcing loadData to be called inline rather than rely on the event firing as it's suppose to. This led me down the rabbit hole of seeing geojson.features return undefined and thinking it must be related to getGeoJSON() not retrieving the features by the time that code executed.

katydecorah:
@riastrad do you have any suggestions why a map that load's a style via `styleLayer` would not allow `map.on('ready')` to be called? I created a reduced test case: https://codepen.io/katydecorah/pen/1b7cc5c4bd03b06b985ed2f49b6df543?editors=0011

I think this is related to https://github.com/mapbox/mapbox.js/issues/1327 and is possibly the reason why the data in this example is getting blocked from rendering.

riastrad:
Thanks for the tag in, @katydecorah. Unfortunately, I don't think that this is related to https://github.com/mapbox/mapbox.js/issues/1327 at all, unfortunately as that's some warning logging that's occurring because https://github.com/mapbox/mapbox.js/pull/1317 didn't account for how `styleLayer` sets a tileJSON property. 

If I change your minimal reproduction to using `v2.3.0` (when `styleLayer` was [first introduced](https://github.com/mapbox/mapbox.js/releases/tag/v2.3.0)) I get the same results, so I think that this is a bug that's been lurking in the `L.mapbox.styleLayer` object since it was implemented.

I think @malwoodsantoro might be on to something here:
> thinking it must be related to getGeoJSON() not retrieving the features by the time that code executed

But I would suspect that it more has to do with the fact that `L.mapbox.styleLayer` extends L.tileLayer and makes two requests (1 for the style.json and 1 for the tileJSON) before it fires a custom `'ready'` event, which may be mucking up Leaflet's preset tileLayer events:
https://github.com/mapbox/mapbox.js/blob/6a6db371daa4ac49940948417ad2c09cf51cdcbb/src/style_layer.js#L26-L48

That's as far as I got this morning, but it might be worth poking this custom event logic & seeing if we can produce different results.

katydecorah:
Thanks for the insight @riastrad!

@malwoodsantoro The problem with using `setTimeout` is that time is relative and the timer is optimize for a user who has a reliable and fast connection. For example, if I disable my cache and set my connection to slow 3G the data never loads because it takes much more time to download the initial data file.

Ideally, the example waits for the data to load before we kick off the `loadData` function by using `on` or some other similar method. (I'm trying to remember if there are other methods in Mapbox.js we could use to make sure `usLayer.getGeoJSON()` is loaded before running `loadData`... I'll keep investigating, but wanted to keep this thread warm.)

katydecorah:
@malwoodsantoro I did some looking and found this example: https://github.com/mapbox/mapbox.js/blob/publisher-production/docs/_posts/examples/v1.0.0/0100-01-01-show-marker-as-label.html#L31-L34 

Here are some suggestions:

1. Rework `var usLayer = .....addTo(map);` to use the pattern in the example above, where the data is loaded through `featureLayer('us.gejoson')` and  in `on` the function then calls `loadData(this.getGeoJSON())`
2. We pass usLayer's geojson to loadData, example: `function loadData(geojson)`.
3. Then `loadData` passes usLayer's `geojson` to joinData, example: `joinData(data, geojson)`

malwoodsantoro:
@katydecorah thanks for the suggestions! Believe it is working as expected now if you wouldn't mind taking another look 🎉 

fix3

docs/_posts/examples/v1.0.0/0100-01-01-choropleth-joined-data-multiple-variables.html
```diff
@@ -73,18 +73,18 @@
...
 // http://shancarter.github.io/mr-data-converter/
-function loadData() {
+function loadData(geojson) {
     $.getJSON('{{site.baseurl}}/assets/data/censusdata.json')
-        .done(function(data) {
-            joinData(data, usLayer);
+        .done(function (data) {
+            joinData(data, geojson);
         });
 }
```
katydecorah, line 90: @malwoodsantoro Two small fixes :)

We'll also need to update `joinData()` function, since we are passing the `usLayer`'s geojson to the function and no longer need to run `usLayer.getGeoJSON()`.

Looking at line 91, `function joinData(data, layer) {` the second parameter `layer` is no longer used, so we could rename it `usGeoJSON`:

```js
function joinData(data, usGeoJSON) {
```

Since the value of `usGeoJSON` is set as the function's parameter, we can replace lines 93-94 with:

```js
var byState = {};
```

malwoodsantoro: Ah, thanks @katydecorah 👍

katydecorah (approved):
💫nice work @malwoodsantoro!

PR #1329, opened 2020-04-24
Resolve console warnings

### LINKED ISSUE #1327: L.mapbox.styleLayer('mapbox://styles/mapbox/streets-v11') throws browser warning about loading "Studio Classic" styles scheduled for deprecation

rmrice:
Fixes #1327 

### Changes
- Only returns console warning about deprecated Studio Classic styles for non-GL styles
- Fixes type error `t.getTileJSON is not a function`. 

Checked all the manual test pages (that are still working), and the console warnings look like what we'd expect.

I think there was also a regression where the mapbox logo was not being added to maps referencing GL style layers. Now the mapbox logo should be added for all mapbox styles. I tried adding a test for this but haven't been successful. Maybe someone more familiar with this codebase has suggestions :)

cc @mapbox/static-apis @katydecorah

### DIFF
--- a/src/load_tilejson.js
+++ b/src/load_tilejson.js
@@ -7,9 +7,13 @@ var request = require('./request'),
 module.exports = {
     _loadTileJSON: function(_) {
         if (typeof _ === 'string') {
+            var style = _;
             _ = format_url.tileJSON(_, this.options && this.options.accessToken);
             var isGLStyle = _.indexOf('/styles/v1/') !== -1;
-            
+            if (!isGLStyle) {
+                util.warn('Warning: this implementation is loading a Mapbox Studio Classic style (' + style + '). ' +
+                    'Studio Classic styles are scheduled for deprecation: https://blog.mapbox.com/deprecating-studio-classic-styles-c65a744140a6')
+            }
             request(_, L.bind(function(err, json) {
                 if (err) {
                     util.log('could not load TileJSON at ' + _);
--- a/src/map.js
+++ b/src/map.js
@@ -210,7 +210,7 @@ var LMap = L.Map.extend({
         // ensure logo appears even when mapbox layer added after map is initialized
         var mapboxLogoControl = this._mapboxLogoControl.getContainer();
         if (!L.DomUtil.hasClass(mapboxLogoControl, 'mapbox-logo-true')) {
-            var tileJSON = layer.getTileJSON();
+            var tileJSON = layer._tilejson;
             this._mapboxLogoControl._setTileJSON(tileJSON);
         }
 
--- a/src/format_url.js
+++ b/src/format_url.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var config = require('./config'),
-    warn = require('./util').warn,
     version = require('../package.json').version;
 
 module.exports = function(path, accessToken) {
@@ -36,15 +35,12 @@ module.exports.tileJSON = function(urlOrMapID, accessToken) {
             'L.mapbox.styleLayer, not L.mapbox.tileLayer');
     }
 
-    if (urlOrMapID.indexOf('/') !== -1)
-        return urlOrMapID;
+    if (urlOrMapID.indexOf('/') !== -1) return urlOrMapID;
 
     var url;
     if (urlOrMapID in config.TEMPLATE_STYLES) {
-        url = module.exports('/styles/v1/' + config.TEMPLATE_STYLES[urlOrMapID], accessToken); 
+        url = module.exports('/styles/v1/' + config.TEMPLATE_STYLES[urlOrMapID], accessToken);
     } else {
-        warn('Warning: this implementation is loading a Mapbox Studio Classic style (' + urlOrMapID + '). ' +
-            'Studio Classic styles are scheduled for deprecation: https://blog.mapbox.com/deprecating-studio-classic-styles-c65a744140a6');
         url = module.exports('/v4/' + urlOrMapID + '.json', accessToken);
     }
 


src/load_tilejson.js
```diff
@@ -7,9 +7,13 @@ var request = require('./request'),
 module.exports = {
     _loadTileJSON: function(_) {
         if (typeof _ === 'string') {
+            var style = _;
             _ = format_url.tileJSON(_, this.options && this.options.accessToken);
             var isGLStyle = _.indexOf('/styles/v1/') !== -1;
-            
+            if (!isGLStyle) {
```
riastrad, line 13: While moving this check over here resolved the issue of logging an incorrect warning for `L.mapbox.styleLayer` it appears that it now means we will log an incorrect warning if a user initializes a `tileLayer` with a third party url (e.g. `openstreetmap.org`) can we add a protocol check as well to make sure we’re not logging for `http`/`https` requests that aren’t for the Mapbox APIs?

src/map.js
```diff
@@ -210,7 +210,7 @@ var LMap = L.Map.extend({
         // ensure logo appears even when mapbox layer added after map is initialized
         var mapboxLogoControl = this._mapboxLogoControl.getContainer();
         if (!L.DomUtil.hasClass(mapboxLogoControl, 'mapbox-logo-true')) {
-            var tileJSON = layer.getTileJSON();
+            var tileJSON = layer._tilejson;
```
riastrad, line 213: 👍 great catch

riastrad (commented):
Thanks for pulling this together so quickly, @rmrice! I have one concern that should be addressed before we roll out a patch release.

src/load_tilejson.js
rmrice: @riastrad I can add a protocol check, but first can you provide an example implementation so I can replicate? I tried one of the examples in this repo (`external-layers.html`), which initializes the map like:

```
    var map = L.mapbox.map('map');
    map.addLayer(L.tileLayer('https://{s}.tile.stamen.com/toner/{z}/{x}/{y}.png'));
    map.setView([0, 0], 5);
```

And I'm not seeing a warning, I think because this implementation is using `L.tileLayer` (leaflet) rather than `L.mapbox.tileLayer`, so it's not actually hitting mapbox.js code. The example breaks when I use `L.mapbox.tileLayer`.

riastrad: Yeah, if it's added as an `L.tileLayer` object we'll be fine. Digging a bit deeper, `L.mapbox.tileLayer` expects the a url string to point to a tileJson endpoint, so it's unlikely a customer will accidentally ship an implementation that attempts to use it this way & subsequently shows this error to customers.

BUT it will lead any customer that tries to use an `L.mapbox.tileLayer` incorrectly to potentially stumble over this incorrect warning and not notice the important `TileJSON` error: 

image

riastrad: And just to drive the point home a bit further, when I point it at a "custom" tileJSON endpoint I'll still log the error. 

**quick code example**
```js
var map = L.mapbox.map('map');
map.addLayer(L.mapbox.tileLayer('https://cyberb.space/work/not-a-real-endpoint.json'));
map.setView([41.88, -87.62], 10);
```

rmrice: Thanks for the examples. After seeing your example along with the tileLayer [docs](https://docs.mapbox.com/mapbox.js/api/v3.3.0/l-mapbox-tilelayer/),
 I don't think a protocol check will work for this since mapbox tileJSON's can be in a similar uri format, like `https://api.mapbox.com/v3/mapbox.dark.json`, but I'm also not sure how an implementation like that would even work since you're just adding a raw tileJSON to the map without a style.

We could check the tileJSON for mapbox tile endpoints, but that seems like overhead for a console warning for a use case that is unlikely to happen & seems to be incorrect.

riastrad: Good point. Okay, I'm comfortable accepting this risk for `L.mapbox.tileLayer` 👍

riastrad (approved):
Looks sharp! 🔪

src/load_tilejson.js
G-Rath: Hi! The company I work for recently inherited a project that makes heavy use of this library, that I believe might be triggering this false warning behaviour via what appears to be correct usage of the library.

The application was using the deprecated styles, which we moved off, but now we're getting these deprecation warnings after having updated to the latest version of this library.

It's very possible we've missed a step and so do have more work to do, but if that is the case I'm not sure what the work would be since we've already gone over the deprecation documentation :/

This is what one of the warnings is: 

> Warning: this implementation is loading a Mapbox Studio Classic style (https://d1lblla2u4yxkc.cloudfront.net/api/Tileset/Melbourne_Outline). Studio Classic styles are scheduled for deprecation: https://blog.mapbox.com/deprecating-studio-classic-styles-c65a744140a6

The application itself is [here](https://pv-map.apvi.org.au/sunspot/map#/map/18/-38.173363784839346/144.34193730310653), and from what we can tell is loading correctly.

Happy to break this out into its own issue, but wanted to leave a comment here first to try and confirm if this is a bug in our app or the library before taking it any further :)

riastrad: 👋 Hey @G-Rath. That warning does look like the edge case I flagged in my original comment (i.e. we're incorrectly classifying your 3rd party endpoint as a legacy endpoint in some cases). Have you tried adding that tilset to your map as an `L.tileLayer` (rather than an `L.mapbox.tileLayer`) to see if that resolves the warning? 

If not, or if you'd prefer a cleaner solution, you can definitely open up a separate issue for this. But just for your awareness, this repository does not have any dedicated maintainers at the moment so bug fixes and new releases can be a bit sporadic.

PR #1331, opened 2020-07-24
[docs] Set defaults for product metadata

katydecorah:
This PR will add defaults to the configuration to set the default values for `platforms`, `services,` and `platforms`. These values will be used internally.

To do:
- [x] Review by @colleenmcginnis to check that values are accurate.
- [x] Update docs-page-shell (Jekyll version) to handle these values (via `{{page.products | jsonify}} {{page.services | jsonify}} {{page.platforms | jsonify}}`).
- [x] Test on staging.

Ref https://github.com/mapbox/documentation/issues/479


katydecorah:
This branch is on staging.

_config.yml
```diff
@@ -13,3 +13,29 @@ defaultid: 'mapbox.streets'
...
+    values:
+      content_type: "example"
+  - scope:
+      path: "_posts/api"
+    values:
+      content_type: "API"
+  - scope:
+      path: "plugins"
+    values:
+      content_type: "example"
```
colleenmcginnis: Could consider tagging plugins as "reference" since they aren't the same kind of examples, but I don't feel strongly about that.

colleenmcginnis (approved):
Looks good @katydecorah! One lil suggestion that I'll leave up to you. 🐊

_config.yml
katydecorah: Per voice, we are going to remove the content_type on the plugins page.

PR #1345, opened 2020-12-04
Final Audit of Examples for Classic Styles

riastrad:
# Context

Support for Mapbox Classic Styles has been turned off (ref. [this blog post](https://blog.mapbox.com/deprecating-studio-classic-styles-d8892ac38cb4)). In order to make sure that Mapbox.js repos continue to function as expected, this PR reflects a final audit of the documentation's use of these resources.

The results of this audit: 
1. Removal of 10 examples that rely on deprecated APIs
2. Updating 7 examples to no longer rely on deprecated classic styles


docs/_posts/examples/v1.0.0/0100-01-01-markercluster-custom-polygon-appearance.html
```diff
@@ -18,12 +18,12 @@
...
 // backing styleLayer, and then use the featureLayer only for its data.
 var map = L.mapbox.map('map')
-    .setView([40.73, -74.011], 13)
+    .setView([39.656, -97.273], 5)
     .addLayer(L.mapbox.styleLayer('mapbox://styles/mapbox/streets-v11'));
 
 // Since featureLayer is an asynchronous method, we use the `.on('ready'`
 // call to only use its marker data once we know it is actually loaded.
-L.mapbox.featureLayer('examples.map-h61e8o8e').on('ready', function(e) {
+L.mapbox.featureLayer('https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson').on('ready', function(e) {
```
katydecorah: @riastrad I see a few examples are referencing "https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson" - would you mind saving this file to this repository and updating the URLs? This would help make these examples more future friendly in case that file were to be deleted from the GL JS repository.

docs/_posts/examples/v1.0.0/0100-01-01-markercluster-custom-marker-icons.html
```diff
@@ -16,12 +16,12 @@
...
 // We add the backing styleLayer and use the featureLayer for its data.
 var map = L.mapbox.map('map')
-    .setView([40.73, -74.011], 13)
+    .setView([39.656, -97.273], 5)
     .addLayer(L.mapbox.styleLayer('mapbox://styles/mapbox/streets-v11'));
 
 // Since featureLayer is an asynchronous method, we use the `.on('ready'`
 // call to only use its marker data once we know it is actually loaded.
-L.mapbox.featureLayer('examples.map-h61e8o8e').on('ready', function(e) {
+L.mapbox.featureLayer('https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson').on('ready', function(e) {
```
katydecorah: I'm seeing some broken icons on this map (no errors logged to console), could it be related to changing the datasource?

image

katydecorah (commented):
This is looking good @riastrad -- I left a couple line comments with suggestions.

For the deleted examples: do you think any of these need redirects? If an example has an alternative, then we'll set up a redirect for it. If the delete example doesn't have an alternative, then we'll allow it to 404 to signal to the user that we no longer support it. If you do want to add redirects, you can create a file in `docs/_posts/examples/redirects/` to set up the redirect.

docs/_posts/examples/v1.0.0/0100-01-01-markercluster-custom-polygon-appearance.html
riastrad: Thanks very much for the nudge on this. Didn't realize we already had this data in our own assets dir. Fixed in e48d4c54.

docs/_posts/examples/v1.0.0/0100-01-01-markercluster-custom-marker-icons.html
riastrad: Yep, turns out the example relies on some simple style spec properties that weren't available in that data. I've updated to a better data source in b6988728.

docs/_posts/examples/v1.0.0/0100-01-01-markercluster-custom-polygon-appearance.html
riastrad: Er, actually fixed in b6988728.

riastrad:
Thanks for the quick review, @katydecorah.

> do you think any of these need redirects?

All of the deleted examples are being removed because they're built on features of the Legacy Maps API that are no longer available. So no redirects feels appropriate. And in the event that the library is updated in such a way that this functionality is possible again without the API, we can always add back versions of these examples.

katydecorah (approved):
Updated examples look great!

PR #1357, opened 2022-06-13
A banner pointing to Mapbox GL JS

mikelmaron:
Customers continue to be confused when they reach our Mapbox.JS documentation (primarily though searches), when their intention was to find our most current library GL JS.

This PR adds a banner to all Mapbox.JS pages

image

@geografa would appreciate input from Support on the best way to frame this, and feedback on whether this is too heavy a touch or an ok presentation.

### DIFF
--- a/docs/_layouts/pages.html
+++ b/docs/_layouts/pages.html
@@ -90,9 +90,8 @@ examples:
   <div class="pad2y">
     <div class="margin3 col9 keyline-all round space-bottom contain fill-orange">
       <div class="contain pad2 fill-lighten3">
-        <h3>Looking for Mapbox GL JS?</h3>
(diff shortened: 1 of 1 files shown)


mikelmaron:
Updated wording to align with [the glossary page](https://docs.mapbox.com/help/glossary/mapbox-js/)
image

docs/_layouts/pages.html
```diff
@@ -86,6 +86,16 @@
 </div>
 
 <div class='limiter clearfix'>
+
+  <div class="pad2y">
+    <div class="margin3 col9 keyline-all round space-bottom contain fill-orange">
+      <div class="contain pad2 fill-lighten3">
+        <h3>Legacy</h3>
+        <p class="prose">Mapbox.js is no longer in active development. To learn more about our newer mapping tools see <a href="https://docs.mapbox.com/mapbox-gl-js/">Mapbox GL JS.</a></p>
```
geografa, line 94: LGTM. Note, https://docs.mapbox.com/mapbox-gl-js/ now resolves to https://docs.mapbox.com/mapbox-gl-js/guides/. Should we just make this the full URL? @mikelmaron

geografa (approved):

